PR #1, opened 2015-02-22
Ochsner OTE2 Config

haraldradi:
A first set of OTE Prameters that I was able to reverse engineer. Will now start to reverse engineer write operations.

I encountered an unsupported datatype, take a look at the parameters heatpump.cycles and heatpump.hours - the value is some strange 2-digit variation of the BCD type, the actual value = value1 + value2 * 100 + value3 * 10000 + value4 * 1000000

Furthermore the BDA type doesn't work as the weekday is returned as 1-7 instead of 0-6 as expected by that type.


john30:
Great, hanks for your work!
Could you paste some example data for the two messages with the datatype in question?
That makes it easier to see what has to be done.

ebusd-0.5.x/ochsner/ote2.csv
```diff
@@ -0,0 +1,43 @@
+*r,ochsner,,,,15,0621,,,,,,,,,,,,,
+*w,ochsner,,,,15,0621,,,,,,,,,,,,,
+r,temperature,outside,Au�entemperatur,,,,00800040,,,parameter,,,,,,temperature,,,,
+r,temperature,water,Warmwasser Temperatur,,,,00840040,,,parameter,,,,,,temperature,,,,
+r,temperature,buffer,Vorlauftemperatur,,,,00e00040,,,parameter,,,,,,temperature,,,,
+r,heating,status,,,,,77800008,,,parameter,,,,status,,UCH,,,,
```
john30: the common prefix (e.g. heating and ID starting with 77) could be nicer written like this (unless the ID suffix 0008 is of any relevance):
*r,heating,,,,15,0621,77,,,IGN:2,,,,,,,,,,
r,,status,,,,,80,,,parameter,,,,status,,UCH,,,,
r,,outside,,,,,81,,,parameter,,,,,,temperature,,,,
...
*r,boiler,,,,15,0621,79,,,IGN:2,,,,,,,,,,
r,,status,,,,,80,,,parameter,,,,status,,UCH,,,,
...

haraldradi:
here is an example payload

cycles: 50811d0001001b500000
hours: 51811D0002000C3D2200

john30:
so for the cycles, "001b5000" would be the value splitted up in four bytes, right? doesn't seem to be BCD anyway. Are you sure about that?

Could you point me to the individual fields for the hours sample? Which one is supposed to be the weekday?

I'd prefer having the full data sequence, i.e. all the bytes starting with the sending master address. that makes it easier to check the message structure.

haraldradi:
ok, we have a misunderstanding here :)

this is not the date type, but that strange unsupported datatype.

The first two bytes (5081 and 5181) are the parameter ids (03-080 and 03-081).
The next byte is, as far as I figured, the the data type identifier (1D in both cases).
The next byte I suppose to be the unit (00, no unit)
The next byte (01 and 02) is the one I ignored, because I don't fully understand the meaning yet.

The rest seems to be the actual value (001b500000 and 000C3D2200):

If you convert that hex -> dec byte by byte you get

001b500000: 00 27 80 00 00
000C3D2200: 00 12 61 34 00

The expected values are (according to the display on my heat pump):

001b500000: 8027
000C3D2200: 3461

which is the decimal value of the bytes read from right to left (i.e. 00 * 1000000 + 00 * 10000 + 80 * 100 + 27 + 00 / 100).

Now that leftover byte on position 5 (01 and 02) seems to be the index of the decimal comma in the value - hence the 00 / 100

Following that assumption the second value would be 003461,1200 using the same decoding strategy. This would also make sense, i just assume that the fractional part is not displayed on the screen.

I have just observed that 6byte long value (including the decimal comma position), but i could imagine that the length could be arbitrary.

I hope that helps. I don't have the raw data at hands right now, just the payloads from my reverse-engineering spreadsheet.

haraldradi:
The date datatype problem is a different one:

r,master,datetime,,,FE,0700,,temperature,,D2B,,,,time,,BTI,,,,date,,BDA,,,,

is a periodic broadcast that i receive which is decoded fine this way except that the weekday is one day off because sunday is expected to be 0x06 but encoded as 0x07

According to the docs in the wiki HDA would be the datatype to use here. But if i take a look at the sources HDA doesn't appear to be a BCD type (unlike stated in the wiki). So either this is a bug or I would need some kind of HDA+BCD datatype.

```c
	{"BDA", 32, bt_dat,     BCD,          0,         10,         10,    0}, // date with weekday in BCD, 01.01.2000 - 31.12.2099 (0x01,0x01,WW,0x00 - 0x31,0x12,WW,0x99, WW is weekday Mon=0x00 - Sun=0x06)
	{"BDA", 24, bt_dat,     BCD,          0,         10,         10,    0}, // date in BCD, 01.01.2000 - 31.12.2099 (0x01,0x01,0x00 - 0x31,0x12,0x99)
	{"HDA", 32, bt_dat,       0,          0,         10,         10,    0}, // date with weekday, 01.01.2000 - 31.12.2099 (0x01,0x01,WW,0x00 - 0x1f,0x0c,WW,0x63, WW is weekday Mon=0x01 - Sun=0x07))
	{"HDA", 24, bt_dat,       0,          0,         10,         10,    0}, // date, 01.01.2000 - 31.12.2099 (0x01,0x01,0x00 - 0x1f,0x0c,0x63)
```

ebusd-0.5.x/ochsner/ote2.csv
haraldradi: the suffix doesn't seem to be of much relevance (in fact I have already observed the same commands with different suffixes, I just haven't figured out where that suffix comes from). nevertheless if I omit that suffix the request packet has a wrong payload length (this is not a broadcast message) and then the request fails.

john30:
Actually, the week day is skipped in all date types, since it does not give any worthful extra information on a system that can easily calculate the week day on its own.
Thus, you can simply use the BDA type with 32 bits (that's the default), and you should be fine.

john30:
Regarding the weird numeric data type:
oh goodness, did someone really think of such a stupid data type syntax?
that's hard to believe, but if you really need it, then ebusd needs to be extended for that.

ebusd-0.5.x/ochsner/ote2.csv
john30: the trick is in the first line of my suggestion: there is an extra "IGN:2" field added to the defaults line, which will automatically be placed before the individual fields of the actual definition. i.e. the suffix you don't know the meaning of yet, is simply ignored.

john30:
Is "ote2" the ident of the device as shown by the scan command?

ebusd-0.5.x/ochsner/ote2.csv
john30: Hi Harald,
please find attached the revised files in CSV and ODF.
cheeers,
John

Am 23.02.2015 um 19:13 schrieb Harald Radi:

haraldradi:
I guess the same guy came up with this data type that also decided to send type, unit, and value range (min, max) with the payload of each value. I think it is some kind of SQL-like "decimal" type to be able to transmit values of arbitrary length. There is one funny thing though, other big values, like the energy usage is split into two separate int-values (MW/h + kW/h).

Nevertheless, I just wanted to share my findings. I can live with the current solution as I will expose those values via a REST interface and can do some further transformations there.

haraldradi:
OTE is the control unit of ochsner heatpumps (http://www.ochsner.com/de/die-waermepumpe/waermepumpen-regelung/)

2 is the version (the one on the website is already version 3)

john30:
ok, lets leave it as it is for now.
Since a future version of ebusd might only load the files fitting to the ident of the scan command, I'd like to know whether the file name "OTE2" matches the ident part of the scan result.

haraldradi:
```
pi@ochsner ~ $ ebusctl
localhost: scan
done

localhost: scan result
15;TEM AG fÃ¼r Elektronik Intertem Vertriebs AG;22102;0472;0110

localhost:
```

Is this what you are looking for? This looks way too generic in my opinion.

john30:
yes, the "22102" is the identification part of the device.
It is planned for a future version of ebusd, that the configuration 
files are automatically loaded depending on this ID from a scan result.
Maybe we should just wait for other Oechsner users to supply further 
configuration details.

Am 25.02.2015 um 14:16 schrieb Harald Radi:

PR #15, opened 2016-02-03
my VRC430 reports with ID "F4300". added a symlink

filnko:


john30:
The 430f is a little bit different than the 430. It was not added to the repository yet, because no one was using it so far.
I'll post the 430f soon

john30:
See commit 900855a

filnko:
oh, great - thanks!

PR #17, opened 2016-03-06
adapt csv to enable "git-beautifying"

MikiDi:
Added some missing commas in the template files so git can display the csv-files in a more readable format.


john30:
thanks!

PR #25, opened 2016-10-07
Update 75.v81.csv

texel007:
Hi, RoomTemp exists already in 470. Therefore recommended name Hc2RoomTemp. KR Texel


john30:
The numbering is invalid, because a room controller can be adjusted to the desired heat circuit.
Furthermore, the name would have to be changed for all room controllers.
BTW: The messages are unique in their naming only in combination with the circuit name. You should rather avoid using the message name alone (without circuit) in order to keep it distinct.

texel007:
ok. understand. I was not aware about the circuit name logic ... :)

Issue #27, opened 2016-10-21
Chaffoteaux TEM 11111.csv


tikismoke:
Select temp from SRA automatique contol 39° then 49 then 43:
```
ebusctl hex 1506210401870040
0a878001027f0080002700
0a878001027f0080003100
0a878001027f0080002b00
```
So this is corresponding to at least to heating water departure temp selected.
Seems circuit pressure ()current value 1.6bar (hexa:10)
```
ebusctl hex 1506210400970040
0a17800d18ff7f00801000
```
Seems hot water selected temperature ()current value 56°C (hexa:38)
```
ebusctl hex 1506210401840040
0a84000102410028003800
```
Seems month next maintenance ()current value 60 (hexa:3c) (confirm by changing it to 15 in the second result
```
ebusctl hex 1506210409de0040
0ade0400003c0000003c00
0ade0400003c0000000f00
```
Average boiler time in min(code 865 in menu) current value87  (hexa:57)
```
ebusctl hex 1506210402d90040
0a5981040cffff00005700
```
Hour boiler heating (code 860 in menu) current value 11 (hexa:b)
```
ebusctl hex 1506210402d60040
0a56810400ffff00000b00
```
Hour boiler hot water (code 861 in menu) current value 6 (hexa:6)
```
ebusctl hex 1506210402d70040
0a57810400ffff00000600
```
Boiler cycle firering (code 863 in menu) current value 353  (hexa:161)
```
ebusctl hex 1506210402d00040
0a50810400ffff00006101
```
Boiler temp:
current measured start heating (code menu 831) current value 32.5°C (hex:145)
```
ebusctl hex 1506210400870040
0a07800d02ff7f00804501
```
current measured back heating (code menu 832) current value 31.4°C (hex:13a)
```
ebusctl hex 1506210400880040
0a08800d02ff7f00803a01
```
current measured hot water (code menu 833) current value 30.1°C (hex:12d)
```
ebusctl hex 1506210400840040
0a04800d02ff7f00802d01
```
current measured exchanger (code menu 840) current value 56.3°C (hex:233)
```
ebusctl hex 15062104008f0040
0a0f800d02ff7f00803302
```
Fan speed (code menu 822) current value: 34x100rpm (hex:22)
```
ebusctl hex 1506210400a90040
0a29800400ffff00002200
```
Current hot water flow debt (code menu 825) current value 4.9L (hex:31)
```
ebusctl hex 1506210400c30040
0a43800000ff0000093100
```
[...]

PR #28, opened 2016-10-25
Patch Tem (chaffoteaux)

tikismoke:
Adding working files.

Still lots to find but a good start.


tikismoke:
#27

john30:
the ods/csv I've sent to you are not yet in a state ready for committing here, we need to work more on those first and at least remove unnecessary comments before commiting. I'll do that once the state is mature enough.

PR #30, opened 2016-11-03
Corrected name of "C700"

Giermann:
In this order it makes more sense to be PrEnergySumHwc2, as it is named in older 08.bai.HW7401.csv and current bai.308523.inc - or do I miss some relevant information?


john30:
that's right. can you verify that it is indeed the value of the seconds hwc unit?

Giermann:
No... it seems my BAI does not have multiple counters, I can only retrieve values for Hc1 and Hwc1, but at least I can confirm it is different from Hwc1, and the naming of ascending IDs let assume the naming for this.

> $ ebusctl r -f PrEnergySumHwc1
> 955805622
>
> $ ebusctl r -f PrEnergySumHwc2
> 0

john30:
okay, good enough. then the naming is wrong in the mdb...

PR #49, opened 2017-03-13
fix copy/paste: z2Name1,z2Name2 description contained 'zone _1_'

doubleO8:
fix copy/paste error:
z2Name1,z2Name2 description contained 'zone _1_'


john30:
thanks!

PR #80, opened 2017-11-28
Hc2OPMode is R/W

guillaumezin:


guillaumezin:
My bad, seems that when I change this register, it is overwritten, reject-it

PR #98, opened 2018-03-08
Include hcmode for HMU

Giermann:
Minimum support for HMU device was already implemented, but leaves a lot of "unknown MS cmd" messages in log file.
It's not complete at all, but reduces at least the unknown issues for many commonly requested messages from VRCxxx control units.


john30:
thanks @Giermann

PR #108, opened 2018-10-24
Support 0010021888 Vaillant ecoTEC plus 256/5-5

cameleon77:
Better match for 0010021888 Vaillant ecoTEC plus 256/5-5


andig:
Looks like https://github.com/john30/ebusd-configuration/pull/107. This seems to be a good default for newer ecoTec models.

cameleon77:
Exactly, I've done it based on your findings :+1:

john30:
do all these devices share the same HW and SW version?
maybe it would be better to use that as the default selection criteria...

cameleon77:
in my case it is SW=0104 HW=7803
0010021888 Vaillant ecoTEC plus 256/5-5

john30:
hm okay thats totally different from what is described in #106, so lets stick to the selection by product ID

PR #139, opened 2019-08-12
Allow write for the broadcast vdatetime. This allows to set heater clock

samm-git:
I found that it is possible to set heater clock on protherm heater (with Thermolink rc/2 thermostat) by sending vtime broadcast. 

This is example.

In one window i am running `listen` command:

```
broadcast vdatetime = 16:01:50;12.08.2019
broadcast vdatetime = 14:01:50;12.08.2019
broadcast vdatetime = 14:01:50;12.08.2019
broadcast vdatetime = 14:02:20;12.08.2019
```

and on another - `ebusctl w -c broadcast vdatetime '14:01:50;12.08.2019'`

As you could see - after my broadcast time is changed and heater starts to use it normally. So my proposal is to make it `r;w`

### DIFF
--- a/ebusd-2.1.x/de/vaillant/broadcast.csv
+++ b/ebusd-2.1.x/de/vaillant/broadcast.csv
@@ -1,6 +1,7 @@
 # type (r[1-9];w;u),circuit,name,[comment],[QQ],ZZ,PBSB,[ID],field1,part (m;s),type / templates,divider / values,unit,comment,field2,part (m;s),type / templates,divider / values,unit,comment,field3,part (m;s),type / templates,divider / values,unit,comment
 *b,broadcast,,,,"FE","B516",,,,,,,,,,,,,,,,,,,
-b,,vdatetime,Datum/Uhrzeit,,,,"00",time,,BTI,,,,date,,BDA,,,,,,,,,
+*wi,broadcast,,,,"FE","B516",,,,,,,,,,,,,,,,,,,
+b;wi,,vdatetime,Datum/Uhrzeit,,,,"00",time,,BTI,,,,date,,BDA,,,,,,,,,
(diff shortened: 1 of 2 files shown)


john30:
I'm not convinced that it is a good idea to take this to the public. it would easily allow unexperienced users to send a bad date/time and thus make the heater go crazy

Dinth:
Im using f47/Date and f47/Time for setting up Date and Time automatically every midnight. The boiler and controller quickly get out of sync whenever the battery in DCF receiver goes flat. 

https://ebusd.fandom.com/wiki/F47/Date
https://ebusd.fandom.com/wiki/F47/Time

samm-git:
+1, in my case every electricity outage completely stops clock in the boiler, as vendor saved a few cents on RTC, so i am doing a sync using ebusd.

samm-git:
@john30 i did requested changes and rebased to master. Also can confirm that it works very well.

WildRat:
**Be especially careful when using this function.** In some cases, heating systems use a different date coding (in particular of the day of the week) - [https://github.com/john30/ebusd/issues/382](https://github.com/john30/ebusd/issues/382) . This does not interfere when receiving data (since in this case the day of the week is discarded), but does not work when sending data. Check what data format your system is using.

samm-git:
@WildRat thank you, will need to check if it renders day of week correctly.

cameleon77:
+1 this is great @samm-git, I have the same issue finding myself with a cold water in the morning due to an electricity outage and clock not updated...

I can confirm this works for me perfectly (Vaillant ecoTEC plus 256/5-5)

Maybe somebody could clarify:
1. Why we need two similar lines in the config:
```
*b,broadcast,,,,"FE","B516",,,,,,,,,,,,,,,,,,,
*wi,broadcast,,,,"FE","B516",,,,,,,,,,,,,,,,,,,
```

2. What would be the correct MQTT example/topic to set vdatetime?

Issue #38, opened 2016-12-18
Vaillant hcmode no pumpstate in Status01


absalom-muc:
Hi john,

Thank you very much for your great program!

I believe that there might be a bug in [hcmode ](https://github.com/john30/ebusd-configuration/blob/master/ebusd-2.1.x/de/vaillant/hcmode.inc) for Vaillant. The last byte of Status01 is not related to the pumpstate. Instead it represents the 'active burner' (similar to Mode/Betriebsart) with the 3 options (off, heater, water). I tested these options with the following results:

```
Heater=off, WW=off
[update info] update MS cmd: 1008b5110101 / 0960500004441e0000ff
[update notice] update bai Status01 QQ=10: 48.0;40.0;4.000;34.0;15.0;off

Heater=on, WW=off
[update info] update MS cmd: 1008b5110101 / 09282600043a420100ff
[update notice] update bai Status01 QQ=10: 20.0;19.0;4.000;29.0;33.0;on

Heater=off, WW=on
[update info] update MS cmd: 1008b5110101 / 09926e00046e1e0400ff
[update notice] update bai Status01 QQ=10: 73.0;55.0;4.000;55.0;15.0;4
```
 Could you have a look on it?
Thank you.
absalom

andig:
@absalom-muc I can confirm your finding for ecoTEC 146. It's the 3-way vent position we're seing in the last byte imho. What I'm wondering- as it's already decoded as on/off/decimal if it should be all decimals or 4 should be mapped to 'warm water' or similar?

absalom-muc:
So far I've the following understanding:
```
bit0: burner for heater (0/1 : off/on)
bit1: overrun (so far I've only seen in combination with bit0=1 (0/1 : no overrun/overrun) ???
bit2: burner for warm water (0/1 : off/on)
```
I will check some logs more in the next days and will give you further feedback. Especially for bit1 I'm not sure if it is really related to an overrun. Do you know what is overunning?

absalom-muc:
In my point of view it makes sense to map '4' to 'warm water'.
But I'm not sure if it is really related to pumpstate only because it is !=0 only when heating is active.
For bit 1 (overrun) I've no more findings

john30:
you should be able to check the values against bai BMUB51101HwcState/BMUB51101BoilerFlowTemp/BMUB51101ErrorStatus/BMUB51101StorageTemp

absalom-muc:
[...]

PR #172, opened 2020-03-08
vaillant/_templates: add pumpstate=4 (warming the hot water circuit)

zajdee:
This diff adds a new `pumpstate`. State `4` has been observed on Protherm heaters with hot water circuit connected, while the heater warms up the water in the boiler. As soon as the water is warmed up, the heater switches to either state `0` (off) or `1` (warming up the home heating circuit).


andig:
Confirmed here: https://github.com/john30/ebusd-configuration/issues/38#issuecomment-342436841

zajdee:
@andig thanks, I didn't know about that issue. I am not sure about the value of `3`, haven't seen it on my heating unit (yet), so in case someone wonders why I only added `4`, this is the reason.

samm-git:
+1, see it as well

PR #177, opened 2020-04-26
Create 12.pmw.hwc.csv

mucheinz:
time periods for circulation pump - english

PR #178, opened 2020-04-26
Create 12.pmw.hwc.csv

mucheinz:
Zeitfenster Zirkulationspumpe - Deutsch


ebusd-2.x.x/de/vaillant/12.pmw.hwc.csv
```diff
@@ -0,0 +1,20 @@
+# type (r[1-9];w;u),circuit,name,[comment],[QQ],ZZ,PBSB,[ID],field1,part (m/s),datatypes/templates,divider/values,unit,comment,,,
```
john30, line 1: why not simply include timer.inc ?

mucheinz: excellent idea - there is no *.inc file in the ebusd-2-x.x path as of now, just in the ebusd-2-1.x path. Does it need to be created here as well?

john30: you should stick to the 2.1.x path anyway, the 2.x.x path is older and only kept for reference. no changes are going to be accepted in 2.x.x
this seems to be a duplicate of #177 anyway, right?

mucheinz: #177 and #178 were for German and English respectively, but no longer needed here.

mucheinz (commented):
no longer needed here

PR #179, opened 2020-05-01
Created 06.vms.csv

Standarduser:
Added 06.vms.csv, based on 06.pms.csv.
Added all working datapoints and dropped the none-working.

### DIFF
--- a/ebusd-2.1.x/de/vaillant/06.vms.csv
+++ b/ebusd-2.1.x/de/vaillant/06.vms.csv
@@ -1,0 +1,41 @@
+# type (r[1-9];w;u),circuit,name,[comment],[QQ],ZZ,PBSB,[ID],field1,part (m/s),datatypes/templates,divider/values,unit,comment
+#,PMS00,VPM 20 S,0020071488 243,,,,,,,,,,
+*r,,,,,,"B509","0D",,,,,,
+*w,,,,,,"B509","0E",,,,,,
+*wi,#install,,,,,"B509","0E",,,,,,
+# ##### Generell #####,,,,,,,,,,,,,
+r,,Ntc1Sensor,NTC 1 Sensor,,,,"0500",,,tempsensor,,,Temperature in solar flow (from solar collector)
+r,,Ntc2Sensor,NTC 2 Sensor,,,,"0600",,,tempsensor,,,Temperature in solar back flow (to solar collector)
+r;wi,,SolPumpRelay,Relais Solarpumpe,,,,"0F00",,,onoff,,,Solar pump relay: 0: OFF; 1: ON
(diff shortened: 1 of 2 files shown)


ebusd-2.1.x/de/vaillant/06.vms.csv
```diff
@@ -0,0 +1,41 @@
+# type (r[1-9];w;u),circuit,name,[comment],[QQ],ZZ,PBSB,[ID],field1,part (m/s),datatypes/templates,divider/values,unit,comment
```
john30, line 1: thanks for the PR. please also provide an english version

Standarduser:
english version -> done

john30:
generic question: does any of the overlapping messages between 06.vms.csv and 06.pms.csv provide any additional information, i.e. different values?
if not, what's the use of having the same messages again but just with a different component name?

Standarduser:
PMS has too much values with error and is not being loaded for VMS controllers. So the primary use is to get VMS working and secondary reduce error messages.

My intention for this pull request was to load online config files and provide this to others asking in ioBroker and Loxone forums for that.

Maybe you have a better idea for realize the config without duplicate data - this is what I got working without being familiar to how rewrite config files from scratch.

Standarduser:
Any news on my pull request? I would like to switch back to online version but don’t want to miss VMS data

ebusd-2.1.x/de/vaillant/06.vms.csv
```diff
@@ -0,0 +1,41 @@
...
+#,VMS01,VMS01,0020071488 243,,,,,,,,,,
```
john30, line 2: please check the product code and enter it here instead of 0020071488. this is revealed by `ebusctl f -c scan.06 -v id` as 'product=...'

ebusd-2.1.x/en/vaillant/06.vms.csv
```diff
@@ -0,0 +1,41 @@
+# type (r[1-9];w;u),circuit,name,[comment],[QQ],ZZ,PBSB,[ID],field1,part (m/s),datatypes/templates,divider/values,unit,comment
+#,VMS00,VMS20 S,0020071488 243,,,,,,,,,,
```
john30, line 2: same as above and align both with each other

john30 (changes requested):

ebusd-2.1.x/de/vaillant/06.vms.csv
Standarduser: Got this answer:
`scan.06 id = prefix=??;year=??;week=??;product=??????????;supplier=????;counter=??????;suffix=??`

Issue #32, opened 2016-11-14
15.350.csv for newer versions


harpss1ngh:
Hi John, as I may have mentioned before, my VRT350 is a newer software than the 15.350.csv file so the DisplayedRoomTemp and DisplayedHwcStorageTemp commands do not work.

However, the commands in 15.370.csv DO work!

So in 15.350.csv it is this:

[SW>=141]r;ws		DisplayedRoomTemp					6600

but in 15.370.csv it is this:

r		DisplayedRoomTemp	DisplayedRoomTemp				8000

Can we add a line in the 15.350.csv with 8000 as the ID? I know someone else with a VRT350 that needs the line from the 15.370.csv for it to work.

How can we get the SW version so we can have a line for example:

[SW>=143]r		DisplayedRoomTemp	DisplayedRoomTemp				8000

john30:
The "info" command will give you the software/hardware versions of your devices, if you started ebusd with --scanconfig (which is the default since 2.1 at least). According to an earlier post from you, your SW version is 0109. You can add the following line below the defaults:
`*[SW],scan,,,SW,,,,,,,,,`
and then you can do things such as using `[SW>=109]` as prefix for lines that are only valid if the software version is at least 109.

PR #184, opened 2020-06-19
SW=0109, 8000 works

GrumpyMeow:
Hi Jon,
thank you for providing ebusd. It must've taken you a lot of time to implement everything.

Initially i was unable to get the measured room temperature from my [Vaillant calorMatic 350](https://www.vaillant.nl/consument/producten/calormatic-350-2182.html) room thermostat. But following the instructions of [this](https://github.com/john30/ebusd-configuration/issues/32) issue, i was able to retrieve the correct information.

The data of my thermostat is:
```
{
     "MF": {"value": "Vaillant"},
     "ID": {"value": "35000"},
     "SW": {"value": "0109"},
     "HW": {"value": "7102"}
}
```


ebusd-2.1.x/en/vaillant/15.350.csv
```diff
@@ -104,3 +104,4 @@ r;wi,,CMResetCnt,,,,,"5F00",,,UIN,,,
 r;wi,,COMErrorCnt,,,,,"6000",,,UIN,,,
 r;wi,,VariantDKRefreshCnt,,,,,"6100",,,UCH,,,
 [SW>=141]r;ws,,DisplayedRoomTemp,,,,,"6600",,,temp,,,
+[SW<141]r;ws,,DisplayedRoomTemp,,,,,"8000",,,temp,,,
```
john30, line 107: please do the same in the german file too

john30 (changes requested):

PR #186, opened 2020-06-25
corrected Status01 and Status02 parameters

SzymekCRX:
Fields' names were repeated


SzymekCRX:
my mistake - still learning :)

Issue #208, opened 2021-01-17
Missing D.14 PumpPowerDesired


ikavior:
Hello.

I have Vaillant ecoTec Plus 206 with Multimatic 700 and in the ebusd I cant find position for D.014 PumpPowerDesired.

ebusd "vaillant/bai.308523.inc", "vaillant/08.bai.csv"

I found that this position is in the bai.0010015600.inc. How to make it working? :)

Thanks.

john30:
start with finding the product ID of the bai by issuing `ebusctl -f -d -c scan`. that has to be added to the bai.csv

ikavior:
scan.08  = Vaillant;BAI00;0104;7803
scan.08 id = 21;19;44;0010021887;0001;012451;N0
scan.15  = Vaillant;70000;0613;6903
scan.15 id = 21;19;43;0020171316;0082;026204;N4

john30:
duplicate of #210

PR #210, opened 2021-01-18
Update bai.308523.inc [Pump Power Rate Settings]

ikavior:
Added D.14 and D.15

### DIFF
--- a/ebusd-2.1.x/en/vaillant/bai.308523.inc
+++ b/ebusd-2.1.x/en/vaillant/bai.308523.inc
@@ -20,6 +20,8 @@ r,,WP,d.10 Central heating pump,,,,"4400",,,onoff,,,internal heating pump
 r,,extWP,d.11 external pump,,,,"3F00",,,onoff,,,External heating pump
 r,,Storageloadpump,d.12 storage load pump,,,,"9E00",,,percent0,,,tank load pump demand
 r,,CirPump,d.13 Circulation pump,,,,"7B00",,,onoff,,,Hot water circulation pump (via external module)
+r;wi,,PumpPowerDesired,d.14 Desired heating pump power,,,,A100,,,UCH,0=auto;1=53;2=60;3=70;4=85;5=100,%,PWM-Desired central heating pump power,,,
+r,,PumpPower,d.15 Current heating pump power,,,,7300,,,UCH,,,Current central heating pump power,,,
 r,,DCRoomthermostat,d.16 room thermostat 24 V,,,,"0E00",,,onoff,,,External controls heat demand (Clamp 3'-4')
 r,,HwcDemand,d.22 DHW demand,,,,"5800",,,yesno,,,domestic hot water (tapping or tank contact) demand
 r,,HeatingSwitch,d.23 operation mode,,,,"F203",,,onoff,,,Wintermode active


ebusd-2.1.x/en/vaillant/bai.308523.inc
```diff
@@ -20,6 +20,8 @@ r,,WP,d.10 Central heating pump,,,,"4400",,,onoff,,,internal heating pump
 r,,extWP,d.11 external pump,,,,"3F00",,,onoff,,,External heating pump
 r,,Storageloadpump,d.12 storage load pump,,,,"9E00",,,percent0,,,tank load pump demand
 r,,CirPump,d.13 Circulation pump,,,,"7B00",,,onoff,,,Hot water circulation pump (via external module)
+r;wi,,PumpPowerDesired,d.14 Desired heating pump power,,,,A100,,,UCH,0=auto;1=53;2=60;3=70;4=85;5=100,%,PWM-Desired central heating pump power,,,
```
john30, line 23: if this is supposed to be added (instead of just linking the right product [as described here](https://github.com/john30/ebusd-configuration/issues/208#issuecomment-782590416)), then please also adjust the "de" file

john30 (changes requested):

ebusd-2.1.x/en/vaillant/bai.308523.inc
ikavior: scan.08  = Vaillant;BAI00;0104;7803
scan.08 id = 21;19;44;0010021887;0001;012451;N0
scan.15  = Vaillant;70000;0613;6903
scan.15 id = 21;19;43;0020171316;0082;026204;N4

john30:
I'm pretty sure this is a clone of the 0010015600. can you simply insert ";'0010021887'" right after "'0010021888'" in the 08.bai.csv file and check if d.14 and d.15 work then?

ikavior:
> I'm pretty sure this is a clone of the 0010015600. can you simply insert ";'0010021887'" right after "'0010021888'" in the 08.bai.csv file and check if d.14 and d.15 work then?

Thats right, it works.

ikavior:
I did changes as you described above in de and en versions.

Thanks for your work!

CherAlban:
I am having the same problem with an ecoTEC plus VC 206/5-5
scan 08 = Vaillant;BAI00;0105;7803
scan.08 id = 21;20;48;0010021927;0001;008456;N1

I tried adding "'0010021927'" right before "'0010021929'" in the 08.bai.csv file, but with no success.

`ebusctl list`  returns many sensors (now incl. d.14 and d.15), but all with "no data stored"
bai AccessoriesOne = no data stored
bai AccessoriesTwo = no data stored
bai AntiCondensValue = no data stored
bai averageIgnitiontime = no data stored
bai BlockTimeHcMax = no data stored
bai BoilerType = no data stored
bai ChangesDSN = no data stored
...

Any ideas what might be wrong?

mskg:
Did you force read the value from the bus with `-f` ?

CherAlban:
> Did you force read the value from the bus with `-f` ?

I added d.14 and d.15 to my local bai.308523.inc (as proposed above), this works flawlessly on my device.

What do you mean with  `-f`? I thought the option stands for `-f, --foreground`

mskg:
You wrote that you don't see any values in the output of your list command. That's normal, because ebusd only shows values that it has either already read or already seen on the bus. 

You force a value to be read by adding a force switch to the read command. I was referring to this switch.

CherAlban:
I tried again, and now everything works as expected. I had forgotten that I had set `filter-seen = 1` in mqtt-hassio.cfg

All values are reported correctly, so it looks like a 0010021927 is a 0010015600 clone.

PR #243, opened 2021-12-31
Normalise CSV column count ebusd 2.1.x

doubleO8:


john30:
the extra columns are there for making github show the file in columns instead of text mode. if they're missing it stops that. there were some PRs in the past adding the extra commas at the end, this PR would basically do the opposite, thus declined.

doubleO8:
Rendering as _text mode_ in github is caused by differences in the number of columns for "header" and "data" lines.
The changes contained in this PR alter the files in such ways that the column count is the same for each line and so it accidentally fixes instead of "stops that" github rendering glitch :)

So for example `50.v61.mc.csv` compare the current master version 

https://github.com/john30/ebusd-configuration/blob/b0c27eb553067dbc77bf721423b4a80b96fb2640/ebusd-2.1.x/de/vaillant/50.v61.mc.csv

VS altered version in PR

https://github.com/doubleO8/ebusd-configuration/blob/1eafbf1fd88a56300415bfb383543aba8f49235b/ebusd-2.1.x/de/vaillant/50.v61.mc.csv

PR #283, opened 2022-11-17
Enable Sunier Duval Genia Set 8 - MiPro

JosemyAB:


JosemyAB:
I'm sorry, this was a mistake

cibernox:
@JosemyAB mistake or not, I have that exact same saunier duval unit. Were you able to make it work?

JosemyAB:
> @JosemyAB mistake or not, I have that exact same saunier duval unit. Were you able to make it work?

Yes! It works. Im using this adapter https://gitlab.com/fromeijn/ebuzzz-adapter. Currently it is not available to buy. But I can send you a Telegram group in which there are people trying to buy it by themself. 

With it plus Jonh30 daemon I'm able to read a lot of sensors and also to change somethings like the desired temp of the hot water tank. Everything using Home Assistant

cibernox:
I already have an ebuzz adapter that I’m using with my wolf heat recuperator in home assistant, no problems with that, I was referring to the ebusd configuration for the genia set. Yesterday I tried to connected the saunier duval but al least out of the box the messages weren’t recognized.

JosemyAB:
Aham OK. Let me share my setup:
* Rpi zero running raspbian
* Ebuzz adapter  connected to the rpi using a USB wire
* ebusd daemon running as debian service

As you can see, [I have my own ebusd-configuration repo](https://github.com/JosemyAB/ebusd-configuration) with my changes (it is a fork from john's one)
I downloaded it in the rpi to use the files locally

This is my startup EBUSD_OPTS env variable.
EBUSD_OPTS="--scanconfig --pollinterval=10 --accesslevel=* --configpath /home/pi/ebusd/ebusd-configuration/latest/en/ --httpport=8889 --latency=100 --mqttjson --mqttint=/home/pi/ebusd/ebusd-configuration/mqtt-hassio.cfg --mqtthost=***** --mqttport=***** --mqttuser=****** --mqttpass=*****"

If you want to join to the Telegram group in which there are a bunch o people working on understand and improve the CSV files for different Vaillant and Sauner Duval aerotherm models, let me know to share the link in private

cibernox:
I very much do want to join. Also, can you confirm that you can use a single adapter to connect two different devices (in my case it would be the genia set and the wolf CWL300). I suspect so because ebus suggests it’s a “bus”, but I couldn’t find info confirming it.

JosemyAB:
> I very much do want to join. Also, can you confirm that you can use a single adapter to connect two different devices (in my case it would be the genia set and the wolf CWL300). I suspect so because ebus suggests it’s a “bus”, but I couldn’t find info confirming it.

I have no idea about if you can use the same ebus adapter devide for two (or more) machines. I would check the documentation. I use mine only with the Genia Se

PR #343, opened 2023-07-24
Vaillant recoVair 260

kmplngj:
I added the config for Vaillant recoVair 260. It is based on the old pull request https://github.com/john30/ebusd-configuration/pull/102. I did the changes I could do. Perhaps someone could have a look at it. I would be happy if I could help adding this.

PR #364, opened 2023-11-11
Add config for recoVAIR 260/4

ojundt:
I'm sharing my working config for a recoVAIR 260/4. The config from https://github.com/john30/ebusd-configuration/pull/343 didn't fully work for me, especially reading the temperatures but then I also found another config [here](https://www.loxforum.com/forum/faqs-tutorials-howto-s/140-howto-vaillant-recovair-l%C3%BCftungsanlage-anbinden/page5). I merged both configs, added some entries like supply and exhaust air speed and polished it up.

Disclaimer: I've tested it with a recoVAIR 260/4 behind a VR32. It also works for that case when the file is stored as `18.v32.csv` (if the VR32 is located at address 18, adapt accordingly!) instead of `08.recov.csv` but from the loxforum it sounds that it works with the same IDs in both cases, just different target address and circuit name.


Rapha811:
Thanks for this PR!

Is it possible to also enable the "Stoßlüften" or "Intensivlüftung" option? My Recovair 260/4 allows to set the "Intensivlüftung" to "on", which leads to a maximum intensity ventilation for the next two (?) hours.

Thanks!

ojundt:
Since I have the recoVAIR behind a CR32 and sensoCOMFORT VRC 720 controller, the 'Stoßlüften' or ventilation stage option is hidden for me on the recoVAIR itself and only available on the controller. But please feel free to use this CSV as the base and add the necessary entry.

In [this CSV](https://www.loxforum.com/forum/faqs-tutorials-howto-s/140-howto-vaillant-recovair-l%C3%BCftungsanlage-anbinden/page5) there was this line which I couldn't validate and hence didn't include. Maybe you can test it:

```
r;w,,FanStage,Luefterstufe,,,,150e,,,UCH,0=low;1=auto;2=mid;3=high,,
```

And here is something about Stoßlüften: https://github.com/john30/ebusd-configuration/pull/343/files#diff-f53d8d6881f15ea1d3503e3a86f330fbee14f2f30ff7a8c3fa35351ba819fd2bR20

Rapha811:
Sorry for the late response. I finally found time to integrate the recoVAIR into my home automation system.

I can confirm that the line above works if the recoVAIR is not behind a CR32. The same is true for

`r;w,,DesiredTemp,Wunschtemperatur,,,,140e,,,temp,,°C,`

Besides, I managed to find the correct register for the intensive ventilation:

`r;w,,IntensiveVentilation,Intensivlüftung,,,,ca0d,,,UCH,0=aus;1=an,,`

You can add this to your PR, if you want to.

john30:
merge with upstream is missing

piotrtobolski:
First, thank you for the config! It was super helpful while integrating with my recoVAIR 360/4. Works like a charm. With one exception:
AirEarthCollector seem to work opposite, when it is disabled in the unit, shows as "yes" in the ebusd and when set to "on" air-earth collector in the ventilation it shows "no" in via ebusd. Maybe the 0/1 values weren't intuitive for this config. Does it work like that for you as well?

piotrtobolski:
I'm also getting those errors for IntensiveVolumeFlow and ReducedVolumeFlow:
```
2024-06-26 17:06:04.140 [update error] unable to parse poll-read recov IntensiveVolumeFlow from 3108b509030d1e0e / 018c: ERR: invalid position
2024-06-26 17:06:08.817 [mqtt error] decode recov IntensiveVolumeFlow: ERR: invalid position
2024-06-26 17:06:58.130 [update error] unable to parse poll-read recov ReducedVolumeFlow from 3108b509030d1d0e / 013c: ERR: invalid position
2024-06-26 17:07:03.874 [mqtt error] decode recov ReducedVolumeFlow: ERR: invalid position
```

piotrtobolski:
I also noticed the `BypassFlapPosition` is either 0% or 62.5%. I don't think I experienced the partial bypass in the last few days to verify what will be shown then. 62.5*16 is 1000 so I think it is issue with datatype in this row.

john30:
can you share your scan result please @ojundt ?

john30:
reworked+merged in 4d119f81

PR #369, opened 2023-11-27
Update 15.700.csv - added Zone3 timers for heating and cooling

stadid:
Added Zone3 timers for heating and cooling. Successfully tested on VRC700/4.

PR #370, opened 2023-11-27
Update 15.700.csv - added Zone3 timers for heating and cooling (de)

stadid:
Added Zone3 timers for heating and cooling. Successfully tested on VRC700/4.


john30:
see the changes requested in #369

PR #389, opened 2024-01-24
Update bai.308523.inc

HiQual:
... and one for the "de" folder


john30:
reworked and merged in ba4c7491

PR #399, opened 2024-02-21
Update bai.0010003886.inc

jonathanschneider:
Fix comment for d.03 and d.04


john30:
more than just one file affected and subject to translating all of them then, which is handled totally different with typespec

john30:
closed as also included in #401

Issue #409, opened 2024-03-20
Vaillant ecoVIT exclusiv VKK 226/4 0010007510


jabdoa2:
I got a Vaillant ecoVIT exclusiv VKK (which reports as `"MF=Vaillant;ID=BAI00;SW=0703;HW=7401", loaded "vaillant/bai.0010007508.inc" ([PROD='0010007510']), "vaillant/08.bai.csv"`). Most of the stuff works out of the box (including `SetMode`).  My model does not have warm water stuff connected but it should support that. I noticed a few things which are not working as expected:

```
# ebusctl find -r | grep ERR
bai actoSTOREEPreventiveCounter =  (ERR: invalid position for 3108b509030dec00 / 00)
bai DCFTimeDate = 09:10:14;17. (ERR: argument value out of valid range for 3108b509030de500 / 07141009170c0507)
bai ExtStorageModulCon =  (ERR: invalid position for 3108b509030dbf00 / 00)
bai FlowSetPotmeter =  (ERR: invalid position for 3108b509030d1900 / 00)
bai PrEnergySumHc1 =  (ERR: invalid position for 3108b509030df400 / 020000)
bai PumpHwcFlowNumber =  (ERR: invalid position for 3108b509030dc200 / 00)
bai PumpHwcFlowSum =  (ERR: invalid position for 3108b509030dc100 / 00)
bai SHEMaxDeltaHwcFlow =  (ERR: invalid position for 3108b509030dc400 / 00)
bai SHEMaxFlowTemp =  (ERR: invalid position for 3108b509030dc300 / 00)
bai SRTFINSDFeedback =  (ERR: invalid position for 3108b509030d8400 / 00)
bai Status16 =  (ERR: invalid position for 3108b5040116 / 00)
bai Status =  (ERR: invalid position for 3108b5110103 / 00)
```

`FlowSetPotmeter` is actually at `f003` similar to some other models but unlike 0010007508 (ecoVit pneumatic/4) which was loaded here. This works:
```
# ebusctl read -m 1 -def 'r,bai,FlowSetPotmeter,Vorlaufsollwert Poti,,08,b509,0df003,temp,s,D2C,,°C,Vorlaufsollwert Potentiometer'
35.19
```
`DCFTimeDate` is probably broken since my heater does not have a DCF antenna (or maybe even receiver) connected. Bus transfer looks like this:
```
[...]

PR #416, opened 2024-04-19
better support for ecoVIT exclusiv VKK 0010007510

jabdoa2:
Fix `FlowSetPotmeter` and `PrEnergy*` for ecoVIT exclusiv VKK 0010007510. This is mostly a copy of 0010007508 with changes for the two entries:
- Use correct address for `FlowSetPotmeter`
- Allow writes (`wi`) for all `PrEnergy*` settings. Otherwise, they will overflow and you cannot reset them
- Use correct address `PrEnergy*`. Those might as well be broken in "0010007508" as `PrEnergySumHwc2` is missing and `PrEnergySumHwc1` kind of duplicate.

Resolves some of the issues from #409


john30:
reworked+merged in 21e5bbea921005f27a6ba261f348a828aff3a8fa

jabdoa2:
@john30 did you somehow adjust the addresses for `PrEnergy`? I did not manage to identify that part in your commit. Maybe you solved it another way?

john30:
please check the [next](https://ebus.github.io/next/) generated csvs with ebusd 24.1 (or current source) as described there or by using a local clone of https://github.com/eBUS/ebus.github.io and using the "next" folder in there

jabdoa2:
> please check the [next](https://ebus.github.io/next/) generated csvs with ebusd 24.1 (or current source) as described there or by using a local clone of https://github.com/eBUS/ebus.github.io and using the "next" folder in there

Looks right. Thanks!

sr258:
Following up on this — the fix here is present in `master`'s TypeSpec source for `bai.0010015600.inc`:

```
@inherit(r, wi_1)
@ext(0xf5, 0)
model PrEnergySumHc1 { ... }
```

and in `master`'s frozen `archived/en/vaillant/bai.0010015600.inc` / `archived/de/vaillant/bai.0010015600.inc` mirrors (both show `r;wi`).

However, the live production CDN at `https://ebus.github.io/{en,de}/vaillant/bai.0010015600.inc` (what ebusd's default `--configpath` actually resolves to) still serves the pre-fix content, with the write gated behind `install_NOT_YET_CONFIRMED` — a level no `--accesslevel` setting can satisfy:

```
w,,install_NOT_YET_CONFIRMED,PrEnergySumHc1,PrEnergySumCH1,,,b509,0ef500,value,,ULG,,,Maintenance data
```

I also checked `https://ebus.github.io/next/en/vaillant/bai.0010015600.inc` (the path referenced above for verification) — same stale result there.

The `ebus.github.io` repo's submodule pin of `ebusd-configuration` is `a6c2904f` (2026-04-19), which postdates the fix commit (`21e5bbe`, Dec 2024), so the source should be included — but the published legacy CSV for this specific product ID doesn't reflect it. Last "import changes" sync affecting this file was 2026-02-01.

Impact: users of this hardware (`ID=BAI00;SW=0405;HW=9602`, product ID `0010021926` → resolves via `[id_hw=9602]` to `bai.0010015600.inc`) still can't reset the `PrEnergySumHc1` maintenance counter via ebusd/MQTT even though the fix has been merged for over a year, because the live CDN ebusd actually fetches from hasn't picked it up.

I also tried pointing `--configpath` directly at `https://raw.githubusercontent.com/john30/ebusd-configuration/master/archived` (and a jsdelivr mirror of the same) as a workaround, but that only has the raw `.inc` files without the generated index/mapping that `--scanconfig` needs, so it fails to resolve anything.

Could the CDN generation/publish step be re-run, or is there a separate confirmation gate blocking this specific product ID's `PrEnergy*` fields from being promoted to the published feed? `ebus.github.io` itself has issues disabled, so posting here instead.

PR #429, opened 2024-08-17
add RecoVair 275

wedr0wycz:


ebusd-2.1.x/en/vaillant/15.vai00.csv
```diff
@@ -0,0 +1,11 @@
+# type (r[1-9];w;u),circuit,name,[comment],[QQ],ZZ,PBSB,[ID],field1,part (m/s),datatypes/templates,divider/values,unit,comment
+#,recov,RecoVair 275,,,,,,,,,,,
+*r,,,,,,B509,0D,,,,,,
+*w,,,,,,B509,0F,,,,,,
```
john30, line 4: this seems to be wrong and should rather be 0E instead of 0F.
please check

ebusd-2.1.x/en/vaillant/c0.wtw04.csv
```diff
@@ -0,0 +1,19 @@
...
+*w,,,,,,B509,0E,,,,,,
+r;w,,BypassValve,,,,,0a00,,,onoff,,,
+r;w,,FrostProtection,,,,,4700,,,onoff,,,
+r;w,,FlowCorrection,Set volume flow correction,,,,2000,,,D1B,,,
+r;w,,room_temperature,,,,,0000,,,temp,,,
+w,,ventilation,ventilation level,,,,1500,,,UIN,,m³/h, wartosc do wymnozenia przez 3/10
+#r,,VolumeFlowAct,,,,,1500,,,UIN,,,
+####### #####,,,,,,,,,,,,,
+*r,,,,,,B509,29,,,IGN:2,,,
+*w,,,,,,B509,,,,,,,
```
john30, line 14: those writes can't work that way as it should read "B509,0E" instead

john30 (changes requested):

PR #462, opened 2025-01-01
Add support for `basv3` VRC 720/3f

burmistrzak:
All registers have been verified to work with a Vaillant sensoCOMFORT VRC 720/3f, running `SW=0760;HW=7304`.

Basis for this PR is the main `basv.tsp` file, with additional messages borrowed from @jonesPD's  version of `ctlv2.csv`.
Conversion to TSP was done manually.

Additional features (compared to `basv`) include:

- Hybrid manager
- Cooling mode
- Improved `SFMode`
- Corrected `OpMode` set
- Offsets/calibration
- Away mode
- Manual temperature setpoints
- Room humidity
- Timers (zone, hwc, cc, silent mode)
- ESCO block function
- Quick Veto mode

Also huge thanks to @stadid & @chrizzzp for their excellent ground work!

PR #330, opened 2023-04-10
New entries for 08.hmu.csv

JongsmaSimon:
08.hmu.csv (English version) has been finalized. It works for Vaillant Arotherm Plus 125/6 and similar heat pumps. It makes available all entries visible in the VWZ AI+ menu.
Three 'write' entries (when uncommented) enable changes from the Home Assistant UI.
Being: Heatcurve, TargetTempHwc & TargetTempHc
@john30 : I could provide German translation, if needed (and I might have more time than you)..

PR #482, opened 2025-01-27
Add support for VRC720 regulators

burmistrzak:
This PR aims to be the most feature-complete and robust definition for the VRC720 sensoCOMFORT line of system regulators.
It's also entirely written in TypeSpec, and already prepared to take advantage of the upcoming [`range`](https://github.com/john30/ebusd/commit/e795ce604aca700b2350e734a59ebc3ca0cfea4d) validation feature for numeric values.

Follow-up for #462 to keep patches manageable.

**Edit:** If all goes well, we'll be able to extent support to the entire suite of [720-based regulators](https://github.com/john30/ebusd-configuration/pull/482#issuecomment-3730754767) from Vaillant and related brands. 🥳

### DIFF
--- a/src/vaillant/15.basv3.tsp
+++ b/src/vaillant/15.basv3.tsp
@@ -3070,6 +3070,7 @@ namespace Basv3 {
 
     /** timer heating monday write */
     @inherit(w_16)
+    @ext(0, 0)
     model Z3HeatingTimer_MondayWrite {
       value: wTimeSlotWithTemp;
     }


john30:
> @john30 What's your policy regarding register names? I'd prefer to simply copy what is being shown on the system regulator's screen, whenever possible. Thoughts?

I'd rather stick to short names for common things like heat circuit="hc", hot water circuit="hwc" etc as otherwise the names grow awfully and unnecessary long, see here https://github.com/john30/ebusd-configuration#component-type-names

and for ease of use messages revealing a single value only should carry the field name "value".

also I'd rather prevent reinvention of the same, i.e. if a message with a suitable name exists elsewhere: use it and rather later do refactoring overall (as is ongoing with the typespec conversion)

kgeree:
hi @burmistrzak, till now i was using CTLV2 config from jonesPD's repo for my regulator, which is (i think a quite early VRC70) identified as "720" in ebusd and I though I give it a try with this basv3 config. I have to say it works almost flawlessly. :)

Few things I noticed:
1) Can't read Errorhistory, while it shouldn't be empty. I get:
`[mqtt error] read 720 Errorhistory: ERR: end of input reached`

2) Have some small issue with PhoneNumber1-2 amd Z*Name1-2
e.g. for phone number I have set 10 digits, in PhoneNumber1 I can see the first 6 digits of it and PhoneNumber2 unable to parse:
```
2025-01-29 08:54:07.847 [update error] unable to parse poll-read 720 PhoneNumber2 from 3115b52406020000007000 / 09030070003432343400: ERR: invalid position
2025-01-29 08:54:07.847 [mqtt error] decode 720 PhoneNumber2: ERR: invalid position
```

Similar issue for Zone names.

3) `Z*QuickVetoDuration` should have a unit "hours" or "h" as currently its just a number.

4) `DewPointOffset `-> Isn't it the Dewpoint only which is displayed? ...and Offset for it is set elsewhere? (I was not able to check it on the regulator as I'm remote) Makes me a bit confused together with the `Hc*DewPointOffset` values...

All in all I think this could be used as a baseline even for devices identified as "720", "ctlv*", "basv*"?

jonesPD:
> 2) Have some small issue with PhoneNumber1-2 amd Z*Name1-2
> e.g. for phone number I have set 10 digits, in PhoneNumber1 I can see the first 6 digits of it and PhoneNumber2 unable to parse:
> ```
> 2025-01-29 08:54:07.847 [update error] unable to parse poll-read 720 PhoneNumber2 from 3115b52406020000007000 / 09030070003432343400: ERR: invalid position
> 2025-01-29 08:54:07.847 [mqtt error] decode 720 PhoneNumber2: ERR: invalid position
> ```

Afaik each register can only a limited set of characters, so two registers are needed for the whole number. However, if you don't program a long enough number in the regulator, you can't read the second register because it is empty. Have you checked this scenario?


> 4) `DewPointOffset `-> Isn't it the Dewpoint only which is displayed? ...and Offset for it is set elsewhere? (I was not able to check it on the regulator as I'm remote) Makes me a bit confused together with the `Hc*DewPointOffset` values...

You can set a dew point offset for each heating circuit per the definition in the VRC720 manual: the minimum supply flow temperature in cooling mode is determined by the calculated dew point + this offset

kgeree:
> @kgeree Were you able to checkout the `testing` branch? It includes the latest HMU, VWZIO, and VRC 720 configs.

not yet...will try later. However looking at the vwzio, I think I have some more values..not sure if you can utilize something, they're all working for me (its a CSV that still needs the tempate files...)
[76.vwzio.hw5103.csv](https://github.com/user-attachments/files/18607749/76.vwzio.hw5103.csv)

HMU and 720 looks good for the first sight (however I'm using ebusd in r/o mode yet) so can't speak for the write options...

kgeree:
okay...I gave it a try, have the following results:
PowerConsumptionVwz - not found, I guess its the same issue like for the HMU

```
[update error] unable to parse poll-read vwzio PowerConsumptionVwz from 3176b5160114 / 00: ERR: invalid position
[mqtt error] decode vwzio PowerConsumptionVwz: ERR: invalid position
```

and I get "null" for the following:
DiagCondensorOutletTemp
DiagFlowTemp
DiagSystemPressure

(but i think vwzio shall not be discussed in this PR)

kgeree:
> You get `null`, but no errors in log, right?

right, null, but no errors. BTW is there a way to show "null" in HomeAssistant as well? then I'd get "Unknown" only if there is really an error on the ebus layer...

chrizzzp:
@burmistrzak 

>     1. I'll investigate the `Errorhistory` – might have to remove it temporarily.

I didn't find the corresponding entries in the CSVs you generated, but I noticed earlier that the `errorhistory` definition is device-specific. See here: 
https://github.com/john30/ebusd-configuration/pull/330#issuecomment-2544008834

I created different `errors.inc` for the different devices and had to add another template, but I'm sure this can be done more elegantly with conditions...

errors_vr71_ctlv2.inc:
```
# error history for vr_71 and ctlv2
*r,,,,,,"B503","01",,,,,,,,,
r,,errorhistory,Error history,,,,01,index,m,UCH,,,,,,errorhistory2
```


errors_hmu_vwz.inc:
```
# error history for hmu and vwz
*r,,,,,,"B503","01",,,,,,,,,
r,,errorhistory,Error history,,,,01,index,m,UCH,,,,,,errorhistory
```


_templates.csv:
```
time2,VTM,,,time
time3,BTM,,,time
date,HDA:3,,,date
date2,BDA:3,,,date
errorhistory,status;time2;date;error,,,
errorhistory2,status;time3;date2;error,,,
```

kgeree:
> @kgeree AFAIK, _Ventilation boost_ only affects your heating zone. Unfortunately named, I guess. 😅
> 
> > This function switches off the heating for 30 minutes. After this (or as soon as you cancel), it will resume the preset operation mode.
> > Frost protection, domestic hot water preparation and circulation all remain active during ventilation boost.

Thats fine. But still not :)
I mean, its kinda stupid. The way in the app to effectively start the Ventilation boost is select at least one Zone **and** the "Ventilation system" - whatever it switches...
image
If I don't switch the "Ventilation System" as well, its the same when I only switch the Z*FSMode to ventilation via ebusd --> App shows ventilation boost active but nothing happens... 
So some extra switch (SFMode??) is missing somewhere IMO

chrizzzp:
@burmistrzak 

> I‘ll give that a try! `errorhistory` on the VRC720 should be more reliable across heatpump variants, right? The system regulator acts as a kind of proxy, so we don’t have to deal with every new software revision. 😉

This would be the ideal case... I'm not so sure if all errors really arrive at (and are recorded in) the regulator. I remember having a 'missing sensor' error of the HWC sensor (supposed to be connected to the vwz) that did not show up in the VRC720, but only in the vwz... I think this should be checked. So what's the easiest way to generate 'non-critical' error messages? 😉

chrizzzp:
@burmistrzak 

> So both, `VR71` and `CTLv2` share the same _type_ of `errorhistory`, correct?

Correct!

stadid:
> @burmistrzak
> 
> > I‘ll give that a try! `errorhistory` on the VRC720 should be more reliable across heatpump variants, right? The system regulator acts as a kind of proxy, so we don’t have to deal with every new software revision. 😉
> 
> This would be the ideal case... I'm not so sure if all errors really arrive at (and are recorded in) the regulator. I remember having a 'missing sensor' error of the HWC sensor (supposed to be connected to the vwz) that did not show up in the VRC720, but only in the vwz... I think this should be checked. So what's the easiest way to generate 'non-critical' error messages? 😉

@burmistrzak, @chrizzzp 

As to my current understanding regulator do not act as error proxy (in terms of "mirroring" heater(s) errors in some regulator register), nor it stores error history.

It has his own error codes. At least for 720 family I've got the list. See below.
[Error_codes-VRC720(f)-VRT380(f)-VR92(f).xlsx](https://github.com/user-attachments/files/18654936/Error_codes-VRC720.f.-VRT380.f.-VR92.f.xlsx)

stadid:
@burmistrzak 

> I assume, you got those codes from the manual(s), right?

No. Found this info by accident on the Internet. Haven't seen any error codes description in the regulator manuals.
This file seems to be some kind of internal use for the service engineers.

stadid:
@burmistrzak 

> At least for the VRC 720f/3, those error codes, including their description and potential fixes, are listed on page 62 of the manual.

Interesting, I've checked  manual for 720/2 and it has only error description text, not the code number.

> Somewhat related question, if I may:
> Have you ever seen any docs regarding the `b509` block for heatpump units?

Unfortunately, no. 
BTW, take a look at the attached document. 
It's very old and seems to be irrelevant to the subject, but maybe it adds something to the the general understanding of the problem.
  
[Vaillant_ebus-2.pdf](https://github.com/user-attachments/files/18669520/Vaillant_ebus-2.pdf)

stadid:
@burmistrzak 
> I don’t think this is the case, but do system regulators at all request operational/diagnostics data? 🤔

If we talk about energy / fuel consumption, forward/return line temp(where applicable) or current heat generator status code/ error, I believe yes, regulator requesting this data with different frequency across the day. I think @chrizzzp mentioned this already. According to my understanding, regulator by himself do not store this data permanently, only acting as a "buffer" or for displaying this data on his screen. 
I've also seen reports when the native app got the energy consumption statistics from the heater while regulator was not connected to the system.  It meas that Vaillant cloud could get those values directly from equipment without regulator. So, the obvious approach is to sniff data between the internet gateway and the heat pump and compare data with what app is showing. The other approach is to sniff data between the regulator and the heat pump and also try to figure out what is requested /answered.

In case you are asking about some specific parameters of the heat generator my guess is if they are not used/displayed by regulator, most probably they are not requested.

stadid:
@burmistrzak

> @stadid Thanks for the detailed response! Well, besides sniffing traffic, another approach would be to analyze the serviceDIALOG diagnostics software.
> 
> It's _highly likely_ that a database of at least `b509` registers is included with this tool. The [VR 15 (0020247921)](https://www.vaillant.de/documents/download/268025) diagnostics adapter can apparently be ordered/purchased, but I don't know if the actual software is included...

Good approach!
Not sure if they ship it with software and is device config database actual or is it updated from their server.

kgeree:
> @kgeree New version with support for `CurrentError` is now available on the `preview` branch. Can you try this one? Maybe simulate an error by temporarily disconnecting the VR71?

nicely working:
image

kgeree:
> > nicely working
> 
> @kgeree Great! ✌️
> 
> > Any chance it'd work the same way for error history?
> 
> Maybe..? Do me a favor, leave the error active and run the following: `ebusctl hex 15b503020101` Then, clear the error and run it again. Post whatever gets returned by the system regulator.

Seems i get some inconsistent readings..


Error on:
```
hex 15b503020101
08025920080225fd01

hex 15b503020101
08000000000000ffff

hex 15b503020101
08020521060824fd01

hex 15b503020101
08022108031124fd01

hex 15b503020101
08022108031124fd01

hex 15b503020101
08025920080225fd01

hex 15b503020101
08025920080225fd01
```


Error off:
```
hex 15b503020101
08025920080225fd01

hex 15b503020101
08025920080225fd01

hex 15b503020101
08025920080225fd01

hex 15b503020101
08025920080225fd01

hex 15b503020101
08020521060824fd01

hex 15b503020101
08023718080225fd01

hex 15b503020101
08023718080225fd01

hex 15b503020101
08000000000000ffff

hex 15b503020101
08025920080225fd01
```

chrizzzp:
@burmistrzak @kgeree 

I think in order to get proper values for historic entries an index has to be added. 

errors.inc
`r,,errorhistory,Error history,,,,01,index,m,UCH,,,,,,errorhistory2`

_templates.csv:
`errorhistory2,status;time3;date2;error,,,`

e.g.
```
ebusctl r -f -c ctlv2 -i 0 errorhistory
ebusctl r -f -c ctlv2 -i 1 errorhistory
ebusctl r -f -c ctlv2 -i 3 errorhistory
...
```

This would translate to:

```
hex 15b50303010101
hex 15b50303010102
hex 15b50303010103
```
       
In my case (no errors in regulator history) this yields

`0800ffffffffffffff`

kgeree:
> @kgeree Ok, decoding based on currently available definitions yields:
> 
> ```
> NN INDEX STATUS TIME_ DATE____ ERROR    
> 08 02    59     20 08 02 25 fd 01
> ```
> 
> What is shown under _error history_ on the system regulator itself?

image

kgeree:
> @kgeree Thx! Can you please run `ebusctl hex 15b50303010101` and post the results? I need to check if the datatype/length for `error` is correct.

here you go:
```
hex 15b50303010101
08023718080225fd01
```

kgeree:
> @kgeree New build on the `preview` branch now includes `ErrorHistory`, give it a shot! 🙌
> 
> You'll have to manually read history entries, as they aren't polled automatically. E.g.:
> 
> ```
> ebusctl r -f -c basv3 -i <INDEX> ErrorHistory
> ```
> 
> The value of `<INDEX>` can be everything from 0...65534, FYI.


for me index is working up to 9 (10 entries - just as it is shown on the regulator as well)

```
read -f -c 720 -i 9 ErrorHistory
9;resolved;12:56;05.04.2024;RECOVAIR_comm_error

read -f -c 720 -i 10 ErrorHistory
ERR: argument value out of valid range in decode: 10;none;00:00;
```

BTW its called "Fault history" on the regulator :)
do you know how to pass the index to fetch it via mqtt from HA?

kgeree:
> Can you please share the output of `hex 15b5030301010a` on your system?

```
hex 15b5030301010a
08000000000000ffff
```

version: ebusd 24.1.p20241115

chrizzzp:
> On a related note, setting the unit for `Z*QuickVetoDuration` to hours, makes it a non-writable counter in HASS. So we either have to remove the unit again, or switch to an Enum for example.
> @chrizzzp Thoughts?

Yes, this is unfortunate. One could also include the unit 'hours' in the name: e.g. `Z*QuickVetoHours` to circumvent, but an Enum is probably better as the values are restricted (0.5 - 12 in 0.5 increments, default 3).

kgeree:
> Can you please simulate another fault – do not clear it, and post the corresponding history entry as hex?

Is this what you mean?

```
read -f -c 720 -i 0 errorhistory
0;resolved;19:44;10.02.2025;RECOVAIR_comm_error

hex 15b50303010101
08025920080225fd01

read -f -c 720 CurrentErrors
RECOVAIR_comm_error;-;-;-;-
```

It seems it shows as "resolved" even if if the error still persists.

kgeree:
> @kgeree Almost. 😉 Here's a step-by-step:
> 
> 1. Simulate a fault
> 2. Get the latest error history item `hex 15b50303010100`, and the previous one for good measure `hex 15b50303010101`
> 3. Resolve the fault
> 4. Repeat step No. 2
> 5. Post all four responses as hex
> 
> We're particularly interested in this byte here: `08 [02] 59 20 08 02 25 fd 01` It's also possible that the `status` byte is completely meaningless for our purposes...

fault present:
```
hex 15b50303010100
08022509110225fd01

hex 15b50303010101
08024419100225fd01
```

fault resolved
```
hex 15b50303010100
08022509110225fd01

hex 15b50303010101
08024419100225fd01
```
so no change there

chrizzzp:
@burmistrzak 

> Check under _Settings > Control > Energy management > Expert settings > Domestic hot water_.

As I have only temporary 'External DHW' via VR71 (heating circuit configured as external DHW), these settings don't show up for me in the app...

chrizzzp:
@burmistrzak 

> Hmm, you mean _Circuit type: DHW_ on the system regulator?

Correct, 'external DHW' is a rather personal nomenclature to differentiate it from the regular (internal) DHW circuit. It also operates an external valve.

kgeree:
> Energy management

I simply don't even have "Energy management" under Settings -> Control. :)
app version: v2.24.3 (23137) with VR921 GW

kgeree:
> > > Energy management
> > 
> > 
> > I simply don't even have "Energy management" under Settings -> Control. :) app version: v2.24.3 (23137) with VR921 GW
> 
> @kgeree Huh, strange. You do have the option _Heating curve settings_ available, right? Also, is EEBus (requires a compatible PV system) configured?

Heat curve settings I have, which is working badly...I once tried to change heat curve for 1 circuit and it changed for all 3 :|..since that I rather not touch...
EEBus I don't have.

chrizzzp:
@burmistrzak 

I think this would make sense (if the differences between the regulators are clear).

chrizzzp:
@burmistrzak 

Presumably, the currently known registers are working on all units.

According to the [manual ](https://www.vaillant.co.uk/documents/controls/sensocomfort/sensocomfort-vrc720(3)-operating-and-installation-instructions.pdf
) there are a few additional settings for the 'v3' vs. the 'v2' units:

/Installer level/ Installation configuration/Installation/DHW alternative point
/Installer level/ Installation configuration/Installation/Flow temp. corr. max. value

With the latter, I'm not sure if this is `HcxExcessTemp` which is also available via ebus in the 'v2' units (but not in the regulator menu). Could you maybe check this with your 'v3' unit?

I'm not sure about the differences between the 'v1' and 'v2' regulators, though.

kgeree:
> @burmistrzak
> 
> Presumably, the currently known registers are working on all units.
> 
> According to the [manual ](https://www.vaillant.co.uk/documents/controls/sensocomfort/sensocomfort-vrc720(3)-operating-and-installation-instructions.pdf) there are a few additional settings for the 'v3' vs. the 'v2' units:
> 
> /Installer level/ Installation configuration/Installation/DHW alternative point /Installer level/ Installation configuration/Installation/Flow temp. corr. max. value
> 
> With the latter, I'm not sure if this corresponds to `HcxExcessTemp` which is also available via ebus in the 'v2' units (but not in the regulator menu). Could you maybe check this with your 'v3' unit?
> 
> I'm not sure about the differences between the 'v1' and 'v2' regulators, though.

afaik all values working fine for me for the (maybe v1) regulator identified as "720".

chrizzzp:
@burmistrzak 

> I don’t think so.
> Did a quick test with the regulator, but AFAICT nothing of interest came up on the eBus…
> It‘s probably another internal value

OK, could be. Did you change the value in the regulator and then checked e.g. `Hc1ExcessTemp` by force-reading with `ebusctl`? Otherwise I would not expect anything to be send over EBUS directly, apart from the fact that the current desired flow temp (of the heat generator, i.e. `SetMode`...) should be in incremented by the value of `Hc1ExcessTemp`.

chrizzzp:
@burmistrzak 

> The _Upper correction value_ is similar to the _Buffer charge offset_ option on regulators used for **SG-Ready**, but specific to **EEBus** and only available in the app.
> It seems to work exactly as advertised, increasing the max. DHW target temperature for overheating when excess PV energy is available.

Interesting, although from my current understanding the regulator itself does not speak EEBUS, only the gateway (VR921 etc.) does and translates everything to EBUS? So does using EEBUS still allow using SG-Ready or are they using the same regulator register(s)? 

```
*r,,,,,,B524,02000000,,,IGN:4,,,
*w,,,,,,B524,02010000,,,,,,
r;w,,CylinderOffset,,,,,A200,,,calibrationv,,,
```

I maybe wrong (I don't have any EEBUS equipment)...

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @write
+  @auth("install")
+  @base(MF, 0x24, 0x2, 1, 0, 0)
+  model wi_1 {}
+
+  /** Hwc bivalence point */
+  @inherit(ri_1, wi_1)
+  @ext(0x1, 0)
+  model HwcBivalencePoint {
+    @minValue(-20) @maxValue(50) @step(1)
```
john30, line 73: step value of 1 should be omitted as this implicitly is the default anyway (depending on the data type though)

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+    value: tempv;
+  }
+
+  /** continuos heating */
+  @inherit(ri_1, wi_1)
+  @ext(0x2, 0)
+  model ContinuosHeating {
+    @minValue(-26) @maxValue(10) @step(1)
```
john30, line 81: compared to 700.tsp the comment was omitted even though it contains valuable information. any reason for that?

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @ext(0xf, 0)
+  model HybridManager {
+    value: hmmode;
+  }
+
+  /** tariff aux heater */
+  @inherit(ri_1)
+  @ext(0x10, 0)
+  model TariffAuxHeater {
+    value: UIN;
```
john30, line 124: what does this value mean? any unit/value list/limit?

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @ext(0x35, 0)
+  model Time {
+    value: HTI;
+  }
+
+  /** hydraulic scheme */
+  @inherit(ri_1, wi_1)
+  @ext(0x36, 0)
+  model HydraulicScheme {
+    value: UIN;
```
john30, line 249: this has a value range for sure

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  /** dew point offset */
+  @inherit(ri_1, wi_1)
+  @ext(0x3A, 0)
+  model DewPointOffset {
+    @minValue(-10) @maxValue(10) @step(1)
+    value: tempv;
+  }
+
+  /** total solar yield */
+  @inherit(r_1)
```
john30, line 276: the yield values are writable in 700.tsp as someone requested to be able to reset those when needed. maybe useful here as well (installer level I'd suggest)

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @ext(0x3d, 0)
+  model SolarYieldTotal {
+    value: energy4;
+  }
+
+  /** total environmental yield */
+  @inherit(r_1)
+  @ext(0x3e, 0)
+  model EnvYieldTotal {
```
john30, line 285: naming should follow 700.tsp

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @inherit(r_1)
+  @ext(0x9a, 0)
+  model GreenIQ {
+    value: onoff;
+  }
+
+  /** hwc cylinder temperature top */
+  @inherit(r_1)
+  @ext(0x9d, 0)
+  model HwcCylinderTempTop {
```
john30, line 469: Cylinder should be named Storage (everywhere)

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @condition(Basv3.HwcEnabled.value, "1")
+  model HwcReheatingActive {
+    value: yesno;
+  }
+
+  /** hwc flow temperature */
+  @inherit(r_2)
+  @ext(0x8, 0)
+  @condition(Basv3.HwcEnabled.value, "1")
+  model HwcFlowTempTarget {
```
john30, line 568: should be "Desired" instead of "Target", also check other occurrences

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  model Hc1FixedDesiredTemp {
+    @minValue(5) @maxValue(90) @step(1)
+    value: tempv;
+  }
+
+  /** DesiredSetbackTemp heating circuit 1 */
+  @inherit(r_3, w_3)
+  @ext(0xd, 0)
+  @condition(Basv3.Hc1CircuitType.value, "2")
+  model Hc1FixedSetbackTemp {
```
john30, line 719: what does set back here mean?

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @ext(0x14, 0)
+  model Hc1SummerTempLimit {
+    @minValue(10) @maxValue(99) @step(1)
+    value: tempv;
+  }
+
+  /** RoomTempMod heating circuit 1 */
+  @inherit(r_3, w_3)
+  @ext(0x15, 0)
+  model Hc1RoomTempMod {
```
john30, line 785: shoud be ...Mode I guess and the comment "room temperature mode ..."

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  model Hc2CircuitType {
+    value: hctype;
+    @maxLength(1) ign: IGN;
+  }
+
+  /** DesiredReturnTemp heating circuit 2 */
+  @inherit(r_4, w_4)
+  @ext(0x4, 0)
+  @condition(Basv3.Hc1CircuitType.value, "4")
+  model Hc2DesiredReturnTemp {
```
john30, line 858: sure this is return temp and not flow temp? wondering what a controller would do with that one instead of the flow temp - adjust the mixer to have the room cooling down/heating up just to reach the target return temp?

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @ext(0x14, 0)
+  model Hc2SummerTempLimit {
+    @minValue(10) @maxValue(99) @step(1)
+    value: tempv;
+  }
+
+  /** RoomTempMod heating circuit 2 */
+  @inherit(r_4, w_4)
+  @ext(0x15, 0)
+  model Hc2RoomTempMod {
```
john30, line 998: again Mod instead of Mode (check all occurrences)

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @ext(0x13, 0)
+  model Z1RoomZoneMapping {
+    value: zmapping;
+  }
+
+  /** heating room temperature desired zone 1 */
+  @inherit(r_6)
+  @ext(0x14, 0)
+  @condition(Basv3.Hc3CircuitType.value, "1")
+  model Z1HeatingRoomTempTarget {
```
john30, line 1371: if it is a desired temp, why is it not writable at all?

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  }
+
+  // ##### timers #####
+
+  /** default *r */
+  model r_9 {}
+
+  @inherit(r_9)
+  @ext(0xb5, 0x55, 0xa0)
+  model UnknownValue_a0 {
```
john30, line 1922: unknown should be commented out

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  model w_9 {}
+
+  /** hwc timer Configuration */
+  @inherit(r_9)
+  @ext(MF, 0x55, 0xa3, 0, 0x2)
+  @condition(Basv3.HwcEnabled.value, "1")
+  model HwcTimer_Config {
+    /** Configuration */
+    @maxLength(9)
+    value: HEX;
```
john30, line 1945: also long hex with unknown meaning should rather be commented out (several times)

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  @condition(Basv3.HwcEnabled.value, "1")
+  model HwcTimer_Monday {
+    value: rTimeSlotWithoutTemp;
+  }
+
+  /** hwc timer monday write */
+  @inherit(w_9)
+  @ext(0, 0)
+  @condition(Basv3.HwcEnabled.value, "1")
+  model HwcTimer_MondayWrite {
```
john30, line 1969: I guess these have to be reworked to have read+write with the same name as desired (rather a comment to myself @john30 )

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  model Z1CoolingTimer_Config {
+    /** Configuration */
+    @maxLength(9)
+    value: HEX;
+  }
+
+  /** timer amount */
+  @inherit(r_9)
+  @ext(MF, 0x55, 0xa4, 0, 1)
+  @condition(Basv3.Hc1CircuitType.value, "1")
```
john30, line 2502: repeated combined conditions could be written more nicely using a dedicated namespace with the conditions on that one, see e.g. [08.ehp.tsp#L524](https://github.com/john30/ebusd-configuration/blob/49d3d38be36cd253e54ede85aa12c931569d10f6/src/vaillant/08.ehp.tsp#L524)

john30 (changes requested):

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3286 @@
...
+  model w_1 {}
+
+  /** default *wi for user level "install" */
+  @write
+  @auth("install")
+  @base(MF, 0x24, 0x2, 1, 0, 0)
+  model wi_1 {}
+
+  /** Hwc bivalence point */
+  @inherit(ri_1, wi_1)
```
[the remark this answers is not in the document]
chrizzzp: Absolutely, "normal" users should be able to read, installers should be able to write.

src/vaillant/15.basv3.tsp
chrizzzp: Agree, comment `-26=off when the outside temperature falls below this threshold temperature the continuous heating function is started (off <=> function is disabled)` should be included to inform about the -26=off setting.

src/vaillant/15.basv3.tsp
chrizzzp: It's the tariff for the secondary heater (Auxiliary Heater, I assume), compared to the tariff of the primary heater (so-called Hybrid Manager, which chooses the most cost-efficient heating mode for a bivalent heater).
Unit for all tariffs could be any (e.g. cents/kWh), but have to be identical for all tariff settings. 
Lower limit should be 0.

src/vaillant/15.basv3.tsp
chrizzzp: It is read-only as it only shows the current desired temperature which is in effect for the zone at the current time (either set by time-progam or by manual operation mode). 
See here (2nd table figure): https://github.com/stadid/ebusd-configuration/pull/2

However, naming it `Z1HeatingRoomTempTarget` could be misleading. Maybe `z1ActualHeatingRoomTempDesired` is better.

src/vaillant/15.basv3.tsp
chrizzzp: Agree.

src/vaillant/15.basv3.tsp
chrizzzp: values list: 1,2,8,9,10,11,12,13,16
or (as in the original regulator) value range: 1-16

src/vaillant/15.basv3.tsp
chrizzzp: No, 'Mod' is actually abbreviation for 'modulation'. The 'room temperature modulation' corresponds to the german 'Raumaufschaltung'. Again, it's the english Vaillant translation... (see here: https://simulator.vaillant.com/vrc720_3/de/#/simulator).

aleek:
This thread is veeeeeeery long and I'm not planning on reading it all, but, I just wanted you guys to know, that I have VRC 720/2 and I can help.

Here's my full dump:
```$ ebusctl scan result
08;Vaillant;HMU00;0308;0403;21;23;48;0010016687;0001;005179;N1
15;Vaillant;CTLV2;0514;1104;21;23;03;0020260916;0953;011657;N0
76;Vaillant;VWZ00;0308;0403
```

flexoTHERM exclusive 117/4
sensocomfort VRC 720/2
weather station (this VWZ00)

```
$ ebusctl grab result
1008b507010a / 03828202 = 119
1008b5110100 / 096c0117000008000100 = 7038
1008b5160111 / 09010506100634060a06 = 5
1076b5110101 / 09ff2db009ff5f0000ff = 42252
1076b5160111 / 00 = 5
10feb505025c00 = 7087
10feb510020601 = 1425
1008b507020900 / 02f702 = 4238
1008b50702092e / 02f702 = 2
1008b50702092f / 02dd02 = 220
1008b507020930 / 02d302 = 506
1008b507020931 / 02b002 = 492
1008b507020932 / 02c902 = 937
1008b507020933 / 02b902 = 441
1008b507020934 / 02b002 = 55
1008b507020936 / 029e02 = 129
1008b50702093c / 027502 = 52
1008b512020464 / 0101 = 559
1008b5120204ff / 0101 = 874
1008b513020528 / 0101 = 1423
1076b512030f0001 / 0732010000801603 = 14701
1076b512030f0002 / 0702020000801603 = 450
1076b512030f0201 / 07fb020000801603 = 25668
1076b512030f0202 / 07d0020000801603 = 1410
1008b516081000ffff02002132 / 0b0000ff020059337c8b934b = 25
1008b516081002ffff02002133 / 0b0201ff02002133609eae48 = 50
1008b516081003ffff02002130 / 0b0301ff0200ea30c03aeb4a = 50
1076b51009000000ffffff010000 / 0101 = 42244
```

chrizzzp:
> On the other hand, I'm wondering why my vrc 720 is called `ctlv2` and yours `basv3`...

`ctlv2` is VRC720/2 (wired) `basv3` is VRC720/3f (wireless). VRC720/3 (wired) is `ctlv3`.

burmistrzak:
Sorry folks... Between health stuff and my studies, I simply haven't had the bandwidth to work on this PR with the attention it deserves.
Don't worry though, I'll get to it sooner than later and review all of your comments. 🤝

I really don't want to accidentally brick anyone's heat pump in the depths of winter. 🥶

burmistrzak:
Welp, I simply couldn't resist and quickly fixed obvious issues. 🙈

src/vaillant/15.basv3.tsp
burmistrzak: Doesn't it help with visibility though?

src/vaillant/15.basv3.tsp
chrizzzp: Good suggestion, I agree to `Z1EffectiveHeatingRoomTempDesired`.

chrizzzp (approved):

scabrero:
> This includes `basv`, `basv3`, `ctlv2` and `ctlv3`, IIRC.

I am using this branch with a wireless Saunier Duval MiPro Sense, SRC 720f/2. It identifies itself as BASS2:

`[bus notice] scan 15: ;Vaillant;BASS2;0507;1704 `

So far, so good.

burmistrzak:
@john30 Ok, after a bit more dog-feeding, I'm confident that this PR is in a mergeable state. 😊
While not yet perfect, it's more than _good enough_ for 98% of use-cases.

I would love to test this with `micro-ebusd` as well, but this doesn't _seems_ to be possible right now?

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3356 @@
...
+  model Hc1ExtHwcTempDesired {
+    value: tempv;
+  }
+
+  /** DewPointOffset heating circuit 1 */
+  @inherit(r_3, w_3)
+  @ext(0xa, 0)
+  @condition(Basv3.Hc1CircuitType.value, "1")
+  @condition(Basv3.Hc1CoolingEnabled.value, "1")
+  @condition(Basv3.Hc1RoomTempModulation.value, ">=1")
```
scabrero: Is this change correct? It was `@condition(Basv3.Hc1DewPointMonitoring.value, "1")`

scabrero (changes requested):
There is definitely some issue with the new namespaces. Loading the generated csvs produces:

`[main error] error resolving conditions: ERR: element not found, last error: condition Basv3 hc1circuittype: message not found, condition Basv3 hc1circuittype: message not found, condition Basv3 hc1coolingenabled: message not found, condition Basv3 hc1coolingenabled: message not found, condition Basv3 hc2circuittype: message not found, condition Basv3 hc2circuittype: message not found, condition Basv3 hc2coolingenabled: message not found, condition Basv3 hc2coolingenabled: message not found, condition Basv3 hc3circuittype: message not found, condition Basv3 hc3circuittype: message not found, condition Basv3 hc3coolingenabled: message not found, condition Basv3 hc3coolingenabled: message not found, condition Basv3 hwcenabled: message not found, condition Basv3 hwcenabled: message not found`

And the conditions declarations are duplicated:

```
*[Basv3_hwcenabled],Basv3,,hwcenabled,,value
*[hwcenabled],,,hwcenabled,,value
```

src/vaillant/15.basv3.tsp
```diff
@@ -0,0 +1,3356 @@
...
+  @inherit(r_9)
+  @ext(MF, 0x55, 0xa3, 0, 1)
+  model Z1CoolingTimer_Config {
+    /** Configuration */
+    @maxLength(9)
+    value: HEX;
+  }
+
+  @condition(Basv3.Hc1CircuitType.value, "1")
+  @condition(Basv3.Hc1CoolingEnabled.value, "1")
```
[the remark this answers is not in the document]
burmistrzak: Looks like you're correct! @john30's implementation of the `@condition` decorator indeed doesn't work properly with TypeSpec namespaces...

src/vaillant/15.basv3.tsp
burmistrzak: @john30 Tried that [here](https://github.com/john30/ebusd-configuration/pull/482#discussion_r2696163750), but multiple conditions don't seem to work quite right with namespaces. Only one condition gets applied.

src/vaillant/15.basv3.tsp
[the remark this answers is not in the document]
john30: the new typespec lib just published now allows these conditions @burmistrzak

chrizzzp:
@agapitoangel 

> Is OpMode in the current 15.basv.csv (CDN) equivalent to Hc1OPMode from the old 2.1.x files? If so, should writing to 
ebusd/basv/OpMode/set with values off/auto/day/night work to control the heating circuit on VRC720f/2?

The OpMode (in combination with OpModeEffect) definitions from the 15.basv.csv do not work with the VRC720 controllers. They apply to older regulators (VRC700). 

The corresponding definitions for the VRC720-series of regulators (e.g. from the current PR) define the OpMode per zone, separate for heating an cooling:

e.g. Z1HeatingOpMode
      Z1CoolingOpMode

with values: 0=off;1=auto;2=manual

The day/night settings from the VRC700 series don't exist anymore as the time slots for heating/cooling are defined by the timers (mode auto).

PR #503, opened 2025-04-27
Add Support for RecoVair 275

mikopp:
This is a modification of PR 429.

I have a fully working version of RecoVair 275 with its remote, all messages are tested and work and are connected to a home automation system.

* I tested the german version (english is copy paste) so please use that as the master if possible
* Lueftungsvolume is really a RPM value but the remote or system forces specific values and shows cubic meters. So the definition makes that east
* The system sets cooling or bypass and not the inverse heat recovery. So i changed all these values to cooling
* I tested all items of remote control (vai) and can change mode and cooling days via a loxone automation, it works nicely
* I tested the read items in wtw and they all work. 
*  writing works on vai (remote) except for Kuehlung on/off which works but has no impact, I guess it is funneled to the recovair itself.
* writing on wtw (recovar itself) works, however this gets immediately overwritten by the remote control and thus only makes sense if remote control is not there
* As I have a working system I can answer questions and am happy to test things
* John, we can also switch to german ... I'm from Austria


mikopp:
Made some more updates in the german version of 15.vai...didn't adapt english version yet.

PR #504, opened 2025-05-02
Add basic support for b516 energy statistics

filippz:
This PR introduces basic support for "b516" statistics present in some `bai` (possibly only present in HW: 7603) gas boilers and `hmu` heat pumps which don't support `PrEnergyCount*`/`PrEnergySum*` values.
Details are explained in the first post here: https://github.com/john30/ebusd-configuration/issues/490#issue-2883618064 and during the discussions there we came to an agreement that only 14 of most useful values are to be defined, although the script (https://gist.github.com/filippz/c0d04e4d607e27fc6b89d127db0ee941) can generate 10794 values based on our findings.

Testing and comments are welcome.

### DIFF
--- a/src/vaillant/08.bai.tsp
+++ b/src/vaillant/08.bai.tsp
@@ -20,6 +20,7 @@ import "./bai.0010021901_inc.tsp";
 import "./bai.0010021961_inc.tsp";
 import "./bai.0010008045_inc.tsp";
 import "./hcmode_inc.tsp";
+import "./b516_inc.tsp";
 using Ebus;
 using Ebus.Num;
 using Ebus.Dtm;
@@ -381,5 +382,6 @@ namespace Bai {
 
     default: Bai._308523_inc, // final load alternative
     Hcmode_inc,
+    B516_inc,
   }
 }
--- a/src/vaillant/08.hmu.tsp
+++ b/src/vaillant/08.hmu.tsp
@@ -2,6 +2,7 @@ import "@ebusd/ebus-typespec";
 import "./_templates.tsp";
 import "./hcmode_inc.tsp";
 import "./errors_inc.tsp";
+import "./b516_inc.tsp";
 using Ebus;
 using Ebus.Num;
 using Ebus.Dtm;
@@ -1793,5 +1794,6 @@ namespace Hmu {
   union _includes {
     Hcmode_inc,
     Errors_inc,
+    B516_inc,
   }
 }
--- a/src/vaillant/15.ctlv2.tsp
+++ b/src/vaillant/15.ctlv2.tsp
@@ -1,6 +1,7 @@
 import "@ebusd/ebus-typespec";
 import "./_templates.tsp";
 import "./errors_inc.tsp";
+import "./b516_inc.tsp";
 using Ebus;
 using Ebus.Num;
 using Ebus.Dtm;
@@ -2551,5 +2552,6 @@ namespace Ctlv2 {
   /** included parts */
   union _includes {
     Errors_inc,
+    B516_inc,
   }
 }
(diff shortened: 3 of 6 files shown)


chrizzzp:
One should add that the `b516` energy statistics are also available from other (including non-heat-generation) circuits, e.g.

- VRC720 regulator (confirmed for ID=CTLV2;SW=0514;HW=1104)
- mixer units (confirmed for ID=VR_71;SW=0104;HW=0503)
- hydraulic station (confirmed for ID=VWZ00;SW=0522;HW=5103)

Depending on the device functionality, these statistics will report in all cases the electrical power consumption of the individual device (even if standby consumption is low). 

In addition to that, the regulator can also provide the energy statistics for e.g. a solar circuit (if controlled by the regulator) and the `b516` energy statistic of an hydraulic station (if equipped with backup heating device, e.g. immersion heater) will report electrical consumption for the backup heating device (hc/hwc) and total consumption (including also decive standby energy).

It is expected also other devices support the `b516` statistics.

filippz:
I've added appropriate includes to `15.ctlv2.tsp`, `26.vr_71.tsp` and `76.vwz.tsp` based on your experience.
We could try to split `b516_inc.tsp` into more logical files so we can include only appropriate values (for example `Fuel*` values make sense only in `08.bai.tsp`) and do some conditional includes (for example as not all bai support b516 statistics and the ones that do don't support `PrEnergy*` values) but this PR aims to just add **basic support** and get more feedback about b516 supported devices.

filippz:
I've removed `SolarCoolingSystem `based on this comment: https://github.com/john30/ebusd-configuration/issues/490#issuecomment-2848733529

Kolbi:
@john30 can you please have a look into it? And maybe merge?

chrizzzp:
@john30 

> > ```
> > r,,,StatEnergySumYear,,,,b516,1003ffff,source,m,UCH,1=solar;2=environment;3=electricity;4=gas,,the energy source,usage,m,UCH,0=all;2=heat;4=water;5=cool,,the energy usage,ign2,m,UCH,=0,,,year,m,UCH,2,,the year as offset from 2000,ign,,IGN:7,,,,value,,EXP,1000,kWh,
> > ```

BTW usage should be: 0=all;**3=heat**;4=water;5=cool

> 
> I've added that one in order not to having to send some hex codes for getting any of the other values. if you want to use it, you need to hand in the source, usage, and the year, e.g. `read -i "1;2;23" statenergysumyear` for getting solar used for heating in 2023.

I see, a nice example for passing additional message parameters to the read command.
Unfortunately, it does not work. 
```
r -i "1;2;23" StatEnergySumYear
ERR: end of input reached
```
The hex message would be:
15b516081003ffff0103002e

I'm still trying to understand the StatEnergySumYear definition. Does it include the zeros preceding 'source', 'usage' and 'year'?

chrizzzp:
> ah my bad, the input needs another ";", so it should be `read -i "1;2;;23" statenergysumyear`

Thanks, now it works. I had to use the -f option when reading several energy values shortly after another.
 
> > I'm still trying to understand the StatEnergySumYear definition. Does it include the zeros preceding 'source', 'usage' and 'year'?
> 
> which zeroes do you mean?

Sorry, my fault, question is obsolete.
Albeit the UCH-coded values for 'source' and 'usage' are single digit in the command (`read -i "1;2;;23" statenergysumyear`), they are, of course, converted to two digit hex bytes (0x01 and 0x02, respectively) present in the hex message. So, no problem here (wouldn't have worked otherwise and could also be easily confirmed with `f -h  statenergysumyear`).
I think I understood the message definition now.

BTW would adding a definition to retrieve the monthly energy sum in a similar fashion be possible? The third parameter byte (translated to 0x00 in the statenergysumyear hex message)  would have to be set in combination with the fourth byte indicating the half-year. I don't see how to achieve this with a divider, as it was possible with the year byte for the annual energy sum. Would one then have to define discrete values for the month and for the year, which would result in quite complex definitions?

PR #509, opened 2025-05-14
Update _templates.tsp add entry to enum daynight

danielstosch:
I added an entry for the instant load status to the enum Values_daynight because of the error message I got: "Ignoring invalid option received on topic 'ebusd/hwc/Mode', got '4', allowed: night, day, floorpaving"


john30:
rephrased it a bit in c70242b0e1b6d9e2b9416cce04c0f0122a9cd702

PR #521, opened 2025-09-02
Add support for room humidity VRC700

muraus:
Added the suppor to read the actual room humidity provided by the internal sensor of the vrc700

PR #524, opened 2025-09-28
added zone 4 and 5, when having extra vr70 coupled to system

garageboxnv:
I have added 2 extra zone/heating circuit definitions to the vrc700 definition, as my system is extended with another vr70 adding 2 zones
A suggestion for improvement could be to have the zone values be defined once abstract (maybe a template) and refer to this with the last byte in how many zones are defined. (e.g. Z1=0x0, Z2=0x1, Z3=0x2, Z4=0x3,...)
Theoretically a system can be extended with more vr70's

P.S. I like the improvement of templates, it gave me the possibility to understand the messages better


garageboxnv:
Added also humidity in the TSP files for VRC700 as in this pull request was made in csv #521 

Was only able to add the VRC700, I still have 2 VR91's but do not have the humidity yet of those.

PR #532, opened 2025-10-14
Weishaupt in tsp

federicoB:
Hello John,

I rewrote the configuration for Weishaupt from [this .csv file](https://github.com/J0EK3R/ebusd-configuration-weishaupt/blob/master/weishaupt/f6..sc.csv) from [J0EK3R](https://github.com/J0EK3R) to .tsp

I used this [documentation](https://github.com/J0EK3R/ebusd-configuration-weishaupt/blob/master/weishaupt/Documentation/f6.md) to determine the addresses and payload structure, and I was able to test almost all values with my heating system.

This PR only includes the heating controller configuration, since that’s the one I needed. I don’t own other Weishaupt devices.

It works perfectly with my Weishaupt WTC-15 heating system and integrates well with my home automation setup.
Thanks again for the adapter!

### DIFF
--- a/src/weishaupt/f1..hc.tsp
+++ b/src/weishaupt/f1..hc.tsp
@@ -15,15 +15,32 @@ enum OperatingPhase {
   PostIgnitionPurge: 0x09,
 }
 
-/** Process values of the heating controller/burner control unit. */
-// broadcast address
+enum OperationMode {
+  BurnerOff: 0x00,
+  NoAction: 0x01,
+  DomesticHotWater: 0x55,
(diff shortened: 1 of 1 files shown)


federicoB:
@andrzejthebestbeast  no you also need to generate the .csv starting from the .tsp
this is the command to do it:
`npm run compile-en`
See guidelines https://github.com/john30/ebusd-configuration/blob/master/guidelines.md#converting-to-csv

Then configure ebusd to read the configuration from the directory where you generated the .csv files

Issue #10, opened 2016-01-03
ebusd --checkconfig failed on ebusd-configuration/ebusd-2.x.x/de/vaillant


vlovo:
1)  _templates.csv : ebus specification types are misssing  for example  
temp,D2C,,°C,Temperatur

2) some issues with duplicated entries , for example in  26.solsy.hc.csv

r,,Status,Status,,,,03,,,temp;press;press;HEX:4,,,,,,
 
 conflicts with 

r,,Timer.Tuesday,Zeitfenster Dienstag,,,,03,,,timer,,,,,,

john30:
For the 2.x.x files, the "--scanconfig" option is required.
Furthermore, the check has to be performed on the base directory (e.g. "ebusd-2.x.x/de"), not on the subdirectory, as the templates in the subdirectory are inherited from the templates in the base directory.

vlovo:
Tx, I´ll  try that...
If I had a look at the closed issues I would have seen  that immediately .....

PR #564, opened 2026-01-30
Add support for 720-based regulators

burmistrzak:
This PR is a superset of #482, including register definitions for the following regulator models:

- `15.720`
- `15.basv`
- `15.basv2`
- `15.basv3`
- `15.ctls2`
- `15.ctlv2`
- `15.ctlv3`
- `15.bass`

If you have one of the aforementioned devices, feel free to give this unified configuration a try and report back!
There're still a few _unsupported_ devices at the moment, specifically:

- `15.ctlv`*
- `15.ctls`*
- `15.ctls3`
- `15.bass2`
- `15.bass3`

_*might not exist..?_

In case you own one of them, please **comment below** with your setup details!

I also have incorporated (hopefully) _all_ the recent discoveries made by @stadid, @chrizzzp & @jonesPD. 🤝
A fork of the CDN repo with pre-compiled definitions [is available](https://github.com/burmistrzak/ebus.github.io) in [English](https://github.com/burmistrzak/ebus.github.io/tree/dev/dev/en) and [German](https://github.com/burmistrzak/ebus.github.io/tree/dev/dev/de).

### Testing & Validation

1. Clone the CDN fork from https://github.com/burmistrzak/ebus.github.io
2. Checkout `dev` branch (should be default)
3. Set `EBUSD_CONFIGPATH` to `<MOUNTPOINT>/ebus.github.io/dev/en`
4. Set `EBUSD_MQTTVAR` to e.g.:
`filter-direction=r|u|^w,filter-circuit=^bas*$,filter-non-circuit=^hmu*$,filter-name=^*,filter-non-name=^Z2|^Z3|^Hc2|^Hc3` (adjust to match your system)
5. Restart `ebusd` container (or service).
6. Check entities in HA and/or MQTT.
7. Look out for errors and report them below.


burmistrzak:
@scabrero Would you mind testing the [forked CDN repo](https://github.com/burmistrzak/ebus.github.io)?

Clone it and point `ebusd` to the [`dev/en`](https://github.com/burmistrzak/ebus.github.io/tree/main/dev/en) folder. 😊

chrizzzp:
@burmistrzak @d3vi1 Great work! 

> “Unknown MS cmd” noise: missing BASV2 message definitions
> ..
> b524 020000004800 (Unknown0048)
   
This message could represent  the "pumpstate" of the main pump (BuildingCircuitPump) in the HP, although not entirely clarified. See discussion here:

https://github.com/stadid/ebusd-configuration/pull/5#issuecomment-2444099534

Maybe someone could confirm/test this?

chrizzzp:
@d3vi1 (probably off-topic):

> P.S.: My BASV2 also includes SOL00 at adress ec. Since this is covered quite well by the ec.sol.sc.tsp template, I believe that only one change would be requried for it to be automatically discovered, a simple symlink: src/vaillant/ec.sol.tsp->src/vaillant/ec.sol.sc.tsp.

My system also reveals a SOL device (but only when running a full scan):

`ebusctl scan full`

`address ec: slave, scanned "MF=Vaillant;ID=SOL00;SW=0514;HW=1104"`

But the definitions in the ` ec.sol.sc.csv` don't work for me (I do have a solar circuit connected via VR71/FM5 which is currently active because it's used with a water-circulating chimney):

```
hex ecb509030D0400
00
```
(should return collector temperature)
```
hex ecb509030D0600
00
```
(should return status of collector pump)

Did you try them?

burmistrzak:
@chrizzzp I'd like to put the `Solar*` stuff behind a condition as well.
Do you have a suggestion, which register would be suitable to check for the presence of a solar circuit?

I can't test this myself, so I need to rely on folks who _have_ that gear installed. 😊

chrizzzp:
@burmistrzak 

`HcxScreedDryingDay` (r/w):
min:0 (0=off)
max: 29
step: 1

`HcxScreedDryingDesiredTemp` (read-only):
min: 0
max: 90
step: 1

In the regulator UI there is a 'screed drying profile' option, which assigns the screed drying temperature (0-90°C, step 1°C) to the different days (1-29). There is also a default schedule, but the corresponding 'schedule' registers have not been found. As far as I remember writing `HcxScreedDryingDesiredTemp` does not work.

burmistrzak:
@chrizzzp If `b516` works correctly with older 720 variants, sure!

While I take care of that, can you please check if there're any other registers in the definition that need _installer_ auth?

burmistrzak:
@d3vi1 Interesting theory!
I unfortunately don't have a second circuit to test this, but maybe someone else does?

But I agree with @stadid that there's no _official_ UI option to change zone/circuit assignments.

burmistrzak:
> All in Wh (Watt hours).

@chrizzzp Huh, seems to be an outlier then. Everything else is in **kWh**.
Should we convert and keep everything in kWh?

burmistrzak:
Also, does anyone know a good way to determine whether a VR92 is available? Ideally, how many are configured as well. 😅
Don't have one, so need your input here.

d3vi1:
> Also, does anyone know a good way to determine whether a VR92 is available? Ideally, how many are configured as well. 😅 Don't have one, so need your input here.

I believe that when we talk to the regulator on ebus, we actually talk to the "Radio receiver unit". If you look at the screen of the regulator you see the following ebus components:
Control (08.05)
Radio receiver unit (05.07)
Heat generator 1 (08.06)
(...)
If you were to look at what ebusd sees, you have:
address 03: master #11
address 04: slave #25, scanned "MF=Vaillant;ID=NETX3;SW=0128;HW=0404", loaded "vaillant/04.netx3.csv"
address 08: slave #11, scanned "MF=Vaillant;ID=BAI00;SW=0806;HW=7603", loaded "vaillant/08.bai.csv"
address 10: master #2
address 15: slave #2, scanned "MF=Vaillant;ID=BASV2;SW=0507;HW=1704", loaded "vaillant/15.basv.csv"
address 26: slave, scanned "MF=Vaillant;ID=VR_71;SW=0100;HW=5904", loaded "vaillant/26.vr_71.csv"
address 31: master #8, ebusd
address 36: slave #8, ebusd
address f1: master #10
address f6: slave #10, scanned "MF=Vaillant;ID=NETX3;SW=0128;HW=0404", loaded "vaillant/f6.netx3.csv"
address ff: master #25

So the BASV2, SW=0507 is actually presented on the bus by the Radio receiver unit. I also believe that it has some sort of register windows to see whatever is hiding behind it (VRC720f/2, VR92f), etc. And I am actively digging into this. I want to extract for each slot of the radio unit the list of SN, PN, ID, HW, MF, etc. This actually might explain why some devices show up twice on the bus (for example SensoNET and NETX3 gateways).

burmistrzak:
@chrizzzp I _think_ we can use `HydraulicScheme` to determine which registers should be available/displayed.
Validation via individual scheme plan and/or simulator. Thoughts?

stadid:
@d3vi1 

> So the BASV2, SW=0507 is actually presented on the bus by the Radio receiver unit. I also believe that it has some sort of register windows to see whatever is hiding behind it (VRC720f/2, VR92f), etc.

If we consider reliability of the system control it make no sense to replace physical bus with radio channel which is not so reliable. Just consider what happens if radio communication is interrupted. System just hangs without the controller.

One of the possibilities is that wireless base station working as "controller without display" and wireless part is just external display with the controls.
Also base station connects to VR92F units. For them wireless link reliability is not that critical as system control just lose temperature data from the assigned zone. If my assumptions are correct VR92F units most probably would not be seen from the wired bus, as they  are "virtually" connected to VRC720/f

stadid:
> Also, does anyone know a good way to determine whether a VR92 is available? Ideally, how many are configured as well. 😅
> Don't have one, so need your input here.

I know at least two working options:

First one is to scan the bus for devices with VR_92 id (suitable for wired controller setups)
Second is to iterate through controller all 9 zones corresponding registers responsible for VR92 unit assignment.
If numeric value is > 2 it means that zone assigned to the VR92 unit with specified address
0 = no assignment
1 = VRC720 itself
2 = VR92 addr 1 
....
5 = VR92 addr 4

burmistrzak:
@stadid Thank you!
I checked Vaillant's planning guidelines and up to nine circuits are indeed possible with multiple heat sources (i.e. hybrid systems). Such a setup would require `HydraulicScheme` be set to `16`.

chrizzzp:
@burmistrzak 

> @stadid Would make sense... Has the entire register address space been searched yet?

Well the _entire_ address space probably not, but quite exhaustively leading to many 'unknown' registers as documented by @jonesPD: https://github.com/jonesPD/ebusd-configuration/blob/master/ebusd-2.1.x/en/vaillant/15.ctlv2.csv

Some contain values that don't seem to change during operation but this is what you would expect from configuration settings...

chrizzzp:
@stadid 

> OR
> 
> you need to have in the system only 2 zones / 2 circuits covered by single FM3 module.
> 
> In this case simulator gives opportunity to change FM3 settings
> 
> Try to start simulation with 2 circuits and you get FM3 instead of FM5 (see below)

OK, I see what you mean. Two heating circuits seem to be the only case where you don't need an FM5 (and FM3 is sufficient). As soon as you add a solar circuit it will require a FM5...

scabrero:
> Please run the following read command:
> 
> ```
> ebusctl hex 15b52406020000002e00
> ```

```
00
```

Probably not the right command.

scabrero:
> @scabrero
> 
> What regulator do you have?

SRC 720f/2, identified as `Vaillant;BASS2;507;1704`

> 
> Please try also: `ebusctl hex 15b52406020000002f00`

```
hex 15b52406020000002f00
0602002f000300
```

I also have a SRC921 gateway that polls this register each 1 hour. The value 3 is reported as "moduleConfigurationVR71" by vaillant cloud API. E.g: https://github.com/signalkraft/myPyllant/blob/main/src/myPyllant/tests/data/heatpump_heat_curve/547b5314b17a38830125171aa35b04ad87ef2de6/system.json#L51

I tried to change my FM3 config from 5 to 1 in the regulator itself and the response was the same, so this register must be something else.

Comparing the response to the FM5 one, the first byte might be the number of zones the system can handle. 02 in the case of FM3, 03 in the case of FM5, maybe 05 in the case of FM3+FM5.

burmistrzak:
@d3vi1 Outstanding!
I'll give it a try a soon as I have fixed my server. 😬

In the meantime, would you mind explaining a bit (in plain English) what you've done, how the script interacts with `ebusd` and how to use it?
Just so folks who aren't well versed in Python and/or CS lingo can understand what's going on and follow along. 😊

chrizzzp:
https://github.com/d3vi1/helianthus-vrc-explorer

@d3vi1 Would this work also with BASV2? The current output seems to be mapped to VRC700 registers.

d3vi1:
> https://github.com/d3vi1/helianthus-vrc-explorer
> 
> @d3vi1 Would this work also with BASVx/CTLVx? The current output seems to be mapped to VRC700 registers.

Not ready for public consumption yet, I am still working on it and it has a few bugs. But it is only a few days away from ready.

For example, without prior knowledge, I haven't yet figured out:

* how to choose the correct optype in B524 for reading. Some Groups are 02h and some are 06h, and there are others as well.
* how to find out the number of instances in a group-generic fashion (I have methods for GG=02, or GG=03, but for others I don't know yet.
* how to find out which instances are enabled.

However, I've managed to figure out the best B524 documentation to date.

But this evening I've figured out which registers are config, which are config-limits, which are state.

But this work should work for all VRC7xx controllers. Mine is a BASV2. And by tomorrow morning the index.html file generation should also be included.

This is how it works: once it finds a slave with address 15h, it runs a B524 discovery on it which lists all the available groups and their types (something not documented anywhere, but observed as a discovery mechanism used by VRC940f. It then attempts to find out which instances are available for each instanced group (not all of them are). It uses a small dictionary for prior knowledge on the number of actually active registers for some well-known groups (02h Heating Circuits, 03h Zones). It then asks you which groups to dump and how many instances. So even if you only have 2 active Zones, you can dump all 11 zones supported by the VRC, regardless of the junk they might contain. Since a large scan might take hours, it allows you to customize what you want to scan.

But please try it on your system and create an issue with the JSON file, and maybe also a dump from pyillant. This should help a lot with understanding the structure of the registers.

But in order to keep this PR on track, please use the discussions in that project. It currently only dumps the extended registers, and not the regular ones, but everything that I've seen in all Vaillant interfaces is already there.

chrizzzp:
Tested the timers from the above `15.720.csv`: Confirmed writing works for both

1) the temperature-bound timers (Z1...) 
```
w -c ctlv2 Z1Timer_Friday "0;1;06:10;22:10;21"
empty

r -f Z1Timer_Friday0
06:10;22:10;21.0
```

2) and the temperature-independent timers (tested with  SilentTimer...).
```
w -c ctlv2 SilentTimer_Friday "0;1;01:00;24:00"
empty

r -f SilentTimer_Friday0
01:00;24:00
```

BTW
- the current definitions contain only one time slot/frame per day (e.g. Z1Timer_Friday0). This might not be enough for many use cases (esp. DHW circulation), I suggest to define at least three time slots/frame per day for all timers or to use indexed reading (-i option) to reduce the amount of required definitions, as suggested here: https://github.com/stadid/ebusd-configuration/pull/3#issue-2169827331 as up to 11 time slots can be defined per day
- there is a nomenclature discrepancy concerning time slots / time frames (the timer definitions use time slot, other definitions use time frame: e.g. Z1Timer_Timeframes), BTW Vaillant uses the term 'time period" in the regulator

chrizzzp:
Concerning this PR some HWC-related suggestions:

The following ExtHWC-related definitions should not be bound to the condition `hc1exthwcactive=1`, since hc1exthwcactive=1 is a dynamic runtime status indicator, that turns on only if the external DHW circuit is currently in operation (i.e. producing hot water). 

```
[hc1exthwcactive=1]r,,,Hc1ExtHwcTempDesired,External hot water desired temperature heating circuit 1,,,b524,020002000900,ign,,IGN:4,,,,value,,EXP,,°C,temperature
[hc1exthwcactive=1]w,,,Hc1ExtHwcTempDesired,External hot water desired temperature heating circuit 1,,,b524,020102000900,value,,EXP,,°C,temperature
...
[hc1exthwcactive=1]r,,,Hc1ExtHwcOpMode,External hot water operation mode heating circuit 1,,,b524,020002001300,ign,,IGN:4,,,,value,,UIN,0=off;1=auto;2=manual,,
[hc1exthwcactive=1]w,,,Hc1ExtHwcOpMode,External hot water operation mode heating circuit 1,,,b524,020102001300,value,,UIN,0=off;1=auto;2=manual,,
```

Instead these definitions should be bound the condition: `[hc1circuittype=3]`, which is a static configuration setting. The same applies to the corresponding hc2, hc3 settings.

Another issue: 
My setup involves such a hc with circuit type 3 (=water=DHW=extHWC) although I don't have an original DHW circuit (i.e. I don't use the circuit to actually produce hot water, but to heat up my buffer cylinder for the heating). This means `HwcEnabled` = 0 (permanently off), but  external DHW circuits can still be defined as active.

However, some of the HWC-related basic settings apply also to the 'external DHW circuit' (e.g. `MaxCylinderChargeTime` and and `HwcLockTime`). Unfortunately, these definitions are not available when the condition `[HwcEnabled=1]` is not met. 
So could they be bound also to the condition e.g. [hc1circuittype=3] or just be unconditional?

burmistrzak:
@chrizzzp Just checked, myVaillant allows max. three time slots per day. 👀

chrizzzp:
And here are a few additions to the 15.720.csv (basic regulator settings):

```
*r,,,,,,B524,02000000,,,IGN:4,,,
*w,,,,,,B524,02010000,,,,,,
r;w,,TariffPrimHeaterLow,,,,,1100,,,UIN,,,Primary heater low tariff value
r;w,,TariffPrimHeaterNormal,,,,,1200,,,UIN,, ,Primary heater normal tariff value
r;w,,MultiInputSetting,multifunction input configuration,,,,6A00,,,UIN,0=not connected;1=1x circulation;2=photovoltaics;3=Ext.Cooling,,
r,,EnergyHeatingThisMonthMirror,,,,,C200,,,ULG,,,
r,,EnergyHeatingLastMonthMirror,,,,,C300,,,ULG,,,
r,,EnergyHeatingHWCTotalMirror,,,,,C500,,,ULG,,,
r,,EnergyHeatingHCTotalMirror,,,,,C600,,,ULG,,,
```

stadid:
> @chrizzzp Just checked, myVaillant allows max. three time slots per day. 👀

For 720 series regulators:
max 12 time slots per day for heating
max 3 time slots per day for DHW
max 3 time slots per day for DHW circulation

john30:
> Tested the timers from the above `15.720.csv`: Confirmed writing works for both

thanks for testing!

>     * the current definitions contain only one time slot/frame per day (e.g. Z1Timer_Friday0). This might not be enough for many use cases (esp. DHW circulation), I suggest to define at least three time slots/frame per day for all timers or to use indexed reading (-i option) to reduce the amount of required definitions, as suggested here: [VRC 720 configuration: understanding and managing Heating Zone time slots stadid/ebusd-configuration#3 (comment)](https://github.com/stadid/ebusd-configuration/pull/3#issue-2169827331) as up to 11 time slots can be defined per day

I recognized that during the review and expected that to be intentional. so if it wasn't then these should get rolled back as well I guess

>     * there is a nomenclature discrepancy concerning time slots / time frames (the timer definitions use time slot, other definitions use time frame: e.g. Z1Timer_Timeframes), BTW Vaillant uses the term 'time period" in the regulator

wrt the slotCountWeek template usage I only see "*_Timeframes" in the old tsp and the new, so where do you see the discrepancy exactly?

chrizzzp:
> > * the current definitions contain only one time slot/frame per day (e.g. Z1Timer_Friday0). This might not be enough for many use cases (esp. DHW circulation), I suggest to define at least three time slots/frame per day for all timers or to use indexed reading (-i option) to reduce the amount of required definitions, as suggested here: [VRC 720 configuration: understanding and managing Heating Zone time slots stadid/ebusd-configuration#3 (comment)](https://github.com/stadid/ebusd-configuration/pull/3#issue-2169827331) as up to 11 time slots can be defined per day
> > 
> I recognized that during the review and expected that to be intentional. so if it wasn't then these should get rolled back as well I guess

Yes, roll back to three time slots **or** - maybe more efficient - have a single read time slot predefined (e.g. Monday0) **and** add a definition that allows indexed reading of any number of time slots:

e.g.
```
[hc1circuittype=1]r,,,Z1Timer_Monday,timer heating monday indexed read 1,,,b555,a5000000,index,m,UCH,,,,htm,,HTM,,,,htm_1,,HTM,,,,slottemp,,UIN,10,°C,desired temperature of a slot
```

> > * there is a nomenclature discrepancy concerning time slots / time frames (the timer definitions use time slot, other definitions use time frame: e.g. Z1Timer_Timeframes), BTW Vaillant uses the term 'time period" in the regulator
> 
> wrt the slotCountWeek template usage I only see "*_Timeframes" in the old tsp and the new, so where do you see the discrepancy exactly?

This is what I meant, it's rather a naming inconsistency. Therefore I suggest to rename "`*_Timeframes`" to "`*_Timeslots`".

john30:
@chrizzzp ok I've addressed your suggestions in d58117571fc2361ca7c07120d38cedfdd2909d29, 0d63db735b661c2346937380971bc53c038dc1d3, 10b51edbca846bfc6c38c1b6c9ac4a3d25eb7308, and 74360dd69ca2c0748f2a6a630c48b37ca404f0a9. I've also made TariffAuxHeater writable as it would be unreasonable otherwise (maybe you can confirm that?).

now to your other findings:
> r;w,,MultiInputSetting,multifunction input configuration,,,,6A00,,,UIN,0=not connected;1=1x circulation;2=photovoltaics;3=Ext.Cooling,,

does that depend on some setup, so is this specific for yours e.g. or would this be available always?

> r,,EnergyHeatingThisMonthMirror,,,,,C200,,,ULG,,,
> r,,EnergyHeatingLastMonthMirror,,,,,C300,,,ULG,,,
> r,,EnergyHeatingHWCTotalMirror,,,,,C500,,,ULG,,,
> r,,EnergyHeatingHCTotalMirror,,,,,C600,,,ULG,,,

if these are "mirrors", where would the non-mirrored value be then? is this related to e.g. PrFuelSumHcThisMonth and which unit do these have (probably kWh)?

john30:
added the MultiInputSetting as well and merged to master in 4875c46 
only thing missing are these mirror energy counters (presumably), therefore kept open

chrizzzp:
@john30 

I investigated more into the Tariff-related definitions. To implement it properly we need a few more definitions, e.g. I just discoverd the tariff timer (for dual tariff operations in trivalence hybrid manager mode, actually quite a neat feature as it considers energy cost and heat demand) and the the register that defines the backup heater  type (gas or electrical). I'm currently trying to find the register that specifies single/dual tariff selection, so the tariff timer can be bound to dual tariff as a condition. I will compile my findings here very soon...

chrizzzp:
OK, here are the updated definitons for the tariff settings. Unfortunately, I could not identify the register for single/dual tariff selection. However, I suggest to make all these definitions conditional to:

`[HybridManager=0]` 
(triVAl mode of hybrid manager, definiton already in the config).

EDIT: The following registers were confirmed to be writable.

```
*r,,,,,,B524,02000000,,,IGN:4,,,
*w,,,,,,B524,02010000,,,,,,
r;w,,TariffBackupHeater,,,,,1000,,,UIN,, ,Secondary heater tariff value
r;w,,TariffPrimHeaterLow,,,,,1100,,,UIN,,,Primary heater dual tariff low tariff value
r;w,,TariffPrimHeaterHighSingle,,,,,1200,,,UIN,,,Primary heater single tariff value/dual tariff high value
r;w,,BackupHeaterType,,,,,1300,,,UCH,0=Condensing;1=Non-Condensing;2=Electrical,,in hybrid manager trival mode only
```
According to the VRC720 manual, BackupHeaterType 0 and 1 are specific to gas boilers (Condensing = "Brennwert" in german). The Vaillant manual uses the term "Backup Boiler" instead of "Backup Heater".  Not sure if we aim for consistency here with Vaillant nomenclature.

Here are the Tariff timer definitions: should be conditional to the dual tariff, but as mentioned I didn't find this register yet.

(sorry for the mix of old and new CSV definitions)

```
r,,,TariffTimer_Timeslots,Tariff timer amount,,,b555,a40006,ign,,IGN:1,,,Amount of configured slots per weekday,slotcount,,UCH,,,number of slots on a day,slotcount_1,,UCH,,,number of slots on a day,slotcount_2,,UCH,,,number of slots on a day,slotcount_3,,UCH,,,number of slots on a day,slotcount_4,,UCH,,,number of slots on a day,slotcount_5,,UCH,,,number of slots on a day,slotcount_6,,UCH,,,number of slots on a day,ign_1,,IGN:1,,,
r,,,TariffTimer_Monday0,Tariff timer monday 1,,,b555,a500060000,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Monday,Tariff timer monday,,,b555,a6000600,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
r,,,TariffTimer_Tuesday0,Tariff timer tuesday 1,,,b555,a500060100,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Tuesday,Tariff timer tuesday,,,b555,a6000601,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
r,,,TariffTimer_Wednesday0,Tariff timer wednesday 1,,,b555,a500060200,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Wednesday,Tariff timer wednesday,,,b555,a6000602,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
r,,,TariffTimer_Thursday0,Tariff timer thursday 1,,,b555,a500060300,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Thursday,Tariff timer thursday,,,b555,a6000603,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
r,,,TariffTimer_Friday0,Tariff timer friday 1,,,b555,a500060400,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Friday,Tariff timer friday,,,b555,a6000604,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
r,,,TariffTimer_Saturday0,Tariff timer saturday 1,,,b555,a500060500,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Saturday,Tariff timer saturday,,,b555,a6000605,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
r,,,TariffTimer_Sunday0,Tariff timer sunday 1,,,b555,a500060600,ign,,IGN:1,,,,htm,,HTM,,,,htm_1,,HTM,,,,ign_1,,IGN:2,,,
w,,,TariffTimer_Sunday,Tariff timer sunday,,,b555,a6000606,slotindex,,UCH,,,index of a time slot,slotcount,,UCH,,,number of slots on a day,htm,,HTM,,,,htm_1,,HTM,,,,wtimeslotconst,,U2L,=65535,,constant value for writing a slot without target temperature
```
The manual states up to 12 tariff time slots can be defined, so I suggest to add indexed reading also for this timer (sorry not done yet for the above snippet).

chrizzzp:
@john30 

> ... only thing missing are these mirror energy counters (presumably), therefore kept open

OK, I had to revise the b524 mirrored energy statistics a little. Some were not assigned correctly (which I hadn't checked before) and I found more.

All values are in kWh.

They are mirrored/collected once per day (I think at midnight) from - I assume - the b516 energy statistics (and converted from Wh to kWh). That's why I suggest a new naming with reference to the b516 `Stat*` energy values. 


```
*r,,,,,,B524,02000000,,,IGN:4,,,
*w,,,,,,B524,02010000,,,,,,
r,,StatMirrorElectricEnergyHcLastYear,,,,,C000,,,ULG,,kWh,
r,,StatMirrorElectricEnergyHcThisYear,,,,,C100,,,ULG,,kWh,
r,,StatMirrorElectricEnergySumThisMonth,,,,,C200,,,ULG,,kWh,
r,,StatMirrorElectricEnergySumLastMonth,,,,,C300,,,ULG,,kWh,
r,,StatMirrorElectricEnergySumThisYear,,,,,C400,,,ULG,,kWh,
r,,StatMirrorElectricEnergySumLastYear,,,,,C500,,,ULG,,kWh,
r,,StatMirrorElectricEnergySumTotal,,,,,C600,,,ULG,,kWh,
```
I haven't checked yet if they are writable, but I assume yes.

These values correspond to the energy values that can be queried from the VRC720 regulator UI. There must be more values (including HWC and more HC and probably solar, environmental and fuel stats).

chrizzzp:
Here are the general Tariff definitions in the current CSV format:

```
r,,install,HybridManager,hybrid manager,,,b524,020000000f00,ign,,IGN:4,,,,value,,UCH,0=trivalence;1=bivalence,,hybrid manager mode
w,,install,HybridManager,hybrid manager,,,b524,020100000f00,value,,UCH,0=trivalence;1=bivalence,,hybrid manager mode
r,,,TariffBackupHeater,tariff backup heater,,,b524,020000001000,ign,,IGN:4,,,,value,,UIN,,,backup heater tariff
w,,,TariffBackupHeater,tariff backup heater,,,b524,020100001000,value,,UIN,,,backup heater tariff
r,,,TariffPrimHeaterLow,low tariff main heater,,,b524,020000001100,ign,,IGN:4,,,,value,,UIN,,,Primary heater dual tariff low tariff value
w,,,TariffPrimHeaterLow,low tariff main heater,,,b524,020100001100,value,,UIN,,,Primary heater dual tariff low tariff value
r,,,TariffPrimHeaterHighSingle,high tariff single main heater,,,b524,020000001200,ign,,IGN:4,,,,value,,UIN,,,Primary heater single tariff value dual tariff high value
w,,,TariffPrimHeaterHighSingle,high tariff single main heater,,,b524,020100001200,value,,UIN,,,Primary heater single tariff value dual tariff high value
r,,install,BackupHeaterType,backup heater type,,,b524,020000001300,ign,,IGN:4,,,,value,,UCH,0=condensing;1=non-condensing;2=electric,,backup heater type
w,,install,BackupHeaterType,backup heater type,,,b524,020100001300,value,,UCH,0=condensing;1=non-condensing;2=electric,,backup heater type
```

IMO the tariff cost definitions should be left out of the installer access level, as tariffs for gas/electricity could change from time to time and this is how it is done also in the VRC720 regulator UI.

I also noticed all time slots of the tariff timer will be reset to `00:00;00:00` as soon as the `HybridManager` register is set to 1`(bivalence)`. The tariff timer slots will default to `11:00;13:00` if `HybridManager` is set to 0`(trivalence)`.

legacycode:
I tested the current 720-based config on a live Vaillant setup and wanted to share one more data point.

Device detected on ebus:
- `MF=Vaillant;ID=CTLV0;SW=0312;HW=9103`
- real device: `Vaillant sensoHOME VRT 380`

For local testing, I mapped `15.ctlv0` to `15.ctlv2` and generated the CSV alias `vaillant/15.ctlv0.csv -> 15.ctlv2.csv`.

Result:
- the device is recognized and loaded successfully
- basic operation works
- the new 720-based definition is a clear improvement over the older config

Remaining issues I still see with this device:
- `FM5Config`: `ERR: invalid position`
- `Hc3CircuitType`: `ERR: invalid position`

So at least for this `CTLV0` variant, `ctlv2` looks like a good baseline, but a few fields still seem not to match the actual payload layout.

chrizzzp:
> * `FM5Config`: `ERR: invalid position`
> 
> * `Hc3CircuitType`: `ERR: invalid position`

Isn't this expected for the VRT380 not being able to control more than a single heating circuit? Is the VR71 (FM5) an option at all for this regulator? If not, this would explain the error messages. 

IMO the VRT380 is not a full 720-based regulator (that this PR addesses). However, good to see that many of the definitions work.

stadid:
@d3vi1
If you are interested, you could reach @legacycode for full id string of his VRT380 and ask him to perform full scan using your solution.

Actually, such behavior of CTLV2 config is expected on VRT380 due to the reasons described above by @chrizzzp.
In simple words VRT380 (at least identified as CTLV0) is a VRC720 limited to one zone/circuit only.

burmistrzak:
@john30 FYI, the `micro-ebusd` service still doesn't like this definition:
```
> bunx -p @ebusd/ebus-typespec tsp2ebusd -m <REDACTED> -i src/vaillant/15.ctlv2.tsp
connecting micro-ebusd...
preparing for micro-ebusd <REDACTED>
Error: compilation failed:
error invalid-schema: Schema violation: must match pattern "^ *[a-zA-Z][a-zA-Z0-9]* *(, *[a-zA-Z][a-zA-Z0-9]+ *)*$" (/withAttrs)
  pattern: ^ *[a-zA-Z][a-zA-Z0-9]* *(, *[a-zA-Z][a-zA-Z0-9]+ *)*$
    at run (file://<REDACTED>/ebusd-configuration/node_modules/@ebusd/ebus-typespec/dist/src/tsp2ebusd.js:199:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

Local fork is at `cdc11e18124e49b6638bb743110219f95e5d405f` for testing.

john30:
> @john30 FYI, the `micro-ebusd` service still doesn't like this definition:

you need to update to the latest https://github.com/john30/ebus-typespec version

stgm:
@burmistrzak and probably @john30  Would you consider using the "Noise Reduction" terminology as used in the control module user interface + the Vaillant manuals? Consistency with the official documentation is my reason to argue for Noise Reduction instead of Silent Operation [here](https://github.com/john30/ebusd-configuration/pull/607/changes#r3215257940).

jlrosende:
BASS3 owner here — happy to provide setup details as requested in the PR description.

## Hardware / setup

- Heat pump: **Saunier Duval GeniaSet HA 10-7.2 STB** (Vaillant rebrand, aroTHERM-class HMU)
- Controller: **BASS3**, `SW=0708`, `HW=4304`
- Companion devices on the same bus: `HMU00` (`SW=0129`, `HW=2204`), `VWZIO` (`SW=0102`, `HW=0504`)
- ebusd 26.1.26.1, eBUS Adapter Shield C6 (TCP enhanced), config loaded from the standard CDN: `vaillant/15.bass.csv`

## `ebusctl info` scan section

```
address 03: master #11
address 08: slave #11, scanned "MF=Vaillant;ID=HMU00;SW=0129;HW=2204", loaded "vaillant/08.hmu.csv"
address 10: master #2
address 15: slave #2, scanned "MF=Vaillant;ID=BASS3;SW=0708;HW=4304", loaded "vaillant/15.bass.csv"
address 31: master #8, ebusd
address 36: slave #8, ebusd
address 71: master #9
address 76: slave #9, scanned "MF=Vaillant;ID=VWZIO;SW=0102;HW=0504", loaded "vaillant/76.vwzio.csv"
```

## Decode failures with the current standalone `15.bass.csv`

Registers that fail with `ERR: invalid position` (raw data shown — the trailing byte group after `/` is what the decoder chokes on):

```
bass HolidayTemp     (ERR: invalid position for 3115b52406020000008200 / 00)
bass ManualCooling   (ERR: invalid position for 3115b52406020000000600 / 00)
bass PrEnergySum     (ERR: invalid position for 3115b52406020000005c00 / 00)
bass Z1DayTemp       (ERR: invalid position for 3115b52406020003000700 / 00)
```

Note the pattern: all four QQ=31, ZZ=15, and the **trailing `/ 00`** suggests the device sends a shorter/different-length answer than the template expects. `Z1DayTemp` in particular is the register HA integrations write for zone setpoints, so it would be great to have it decodable.

Registers that decode fine on the same unit (for comparison): `Z1OpMode`, `Z1NightTemp`, `Z1ActualRoomTempDesired`, `Z1QuickVetoTemp`, `HwcOpMode`, `HwcSFMode`, `HwcTempDesired`, `HwcStorageTemp`, `Hc1Status`, `Hc1PumpStatus`, `Hc1FlowTemp`, `Hc1Min/MaxFlowTempDesired`, `Hc1HeatCurve`, `Hc1SummerTempLimit`, `PrEnergySumHc`, `PrEnergySumHwc`, `YieldTotal`, `GlobalSystemOff`, `ContinuosHeating`.

Also on the same bus, `08.hmu.csv` produces `invalid position` for: `YieldTotal`, `ConsumptionTotal`, `Status`, `FlowTemp`, `SourceTempOutput`, `RunDataStatuscode` (this matches issue #624 — same `HMU00;SW=0129`).

## Additional data available

- Full anonymized register dump (871 registers, 241 writable): https://gist.github.com/jlrosende/02b291c1002b5a3e33ac535130a81f7b
- I can capture raw eBUS traffic (the C6 adapter + ebusd `grab`/inject workflow) for any specific register ID if that helps validate the 720-family templates against BASS3 — just tell me which IDs to grab.

Happy to test a pre-compiled `15.bass`/`15.bass3` CSV from the `burmistrzak/ebus.github.io` fork on this hardware and report back a before/after of the four failing registers.

d3vi1:
Thanks for the very detailed BASS3 trace and dump. I cross-checked the four selectors against the Helianthus B524 documentation and register map:

- `ManualCooling`: `02 00 00 00 06 00` is `OP=0x02, GG=0x00, II=0x00, RR=0x0006`. This matches my `manual_cooling_days` register. An empty `/ 00` reply should not be decoded as a differently sized value; it is a non-value response. For this specific register, it is consistent with the documented “dormant” state when cooling is not configured.

- `PrEnergySum`: `02 00 00 00 5c 00` matches my total energy-consumption counter (`RR=0x005c`, normally `u32`). The selector is therefore plausible, but the empty reply contains no value/header to decode. I can't interpret it.

- `Z1DayTemp`: `02 00 03 00 07 00` is a mismatch. In my current map, the Zone 1 heating setpoint is `OP=0x02, GG=0x03, II=0x00, RR=0x0022`; `RR=0x0007` is not that setpoint. This looks more like a BASS3/CSV mapping issue than a scalar decoder-length issue.

- `HolidayTemp`: `02 00 00 00 82 00` is not equivalent to the per-zone holiday setpoint in our map (`OP=0x02, GG=0x03, RR=0x0005`). It may be a separate controller-global BASS3 field, but I do not yet have enough evidence to assign that meaning.

I've asked an agent to create a small gist to allow you to browse the B524 protocol based on my reverse engineering for the emergent Helianthus project. Don't look at B524 as a flat register map. It has states, it has gates (both for hardware and depending on config), it has different operations which behave differently. I've also documented the Helianthus view in parallel with the eBUSd view. If you can identify another register, please contribute as an issue to https://github.com/Project-Helianthus/helianthus-docs-ebus . Also try helianthus-vrc-explorer to see what you can discover on your system and upload your findings to our project. Maybe we can enrich it a bit.

Anyway, hope it helps you understand the registers a little better. Here's my little GIST with almost everything I have in a slightly friendlier fashion than the formal project documentation:
https://gist.github.com/d3vi1/df92aeb6675414cd3eb3e6255f5c3e62

PR #591, opened 2026-04-14
Update 15.ctlv2.tsp adding GlobalSystemOff and HwcEnabled

Tux92:
Previously on CSV file I have this 2 entity, that I use on holliday or if I want to stop Hwc production.
So now I add them to tsp new config.

This changes work with Vaillant aroTHERM plus VWL 85/6 A 230V S3


john30:
this is done in #564

Issue #446, opened 2024-11-19
missing HC1 Returntemp after updating ebus to 24.1


Jpaul1001:
thank you very much for the excellent work!

unfortunately the value for the return temperature of heating circuit 1 is no longer available!?

before: V23.3
ebusd_hmu00_returntemp_temps2
ebusd_hmu00_status01_temp1_2

switschel:
Same here returnTemp is missing in the current & next ebusd configuration. Any reasons it was dropped or not added?

roe-lz:
I'm also looking for an accurate source of returnTemp for vaillant heatpumps to be able to calculate COP. But I'm wondering if there actually is a good source for this?

I think it was available in older csv versions as ebusd_hmu00_returntemp_temps2 or sometimes just returnTemp, but afaik always pointing to b5140529.

b5140529 is still present in the latest and next versions, but is now called TestReturnTemp, which makes sense as it is under the installer test menu of the heat pump controller. 

However, I cannot actively read this value. It is only broadcasted on the ebus when I enter the installer test menu.  

So if someone is able to read accurate returnTemp from the hmu I am very interested :)

roe-lz:
After a lot of testing and trying, I found two addresses that seem to work for me, and are accurate enough for what I want to do (calculate COP).

I added these in 08.hmu.csv:

`r,,,RunDataFlowTemp,,,,b509,540200fc08,value,,IGN:4,,,,value,,EXP,,°C,temperature,
r,,,RunDataReturnTemp,,,,b509,5402000609,value,,IGN:4,,,,value,,EXP,,°C,temperature`

If others can test this on their installation and let me know if it works for them too then I will try to create a PR for these.

Jpaul1001:
Problem solved,
I made the following entry at the end of the 08.hmu.csv:
r1,,ReturnTemp,,,,,2903FFFF,,,temps2,10,°C,Return temperature (°C),,,,,,,,,,,,,,,,,,
!include,hcmode.inc,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,

this reads out the return temperature.

switschel:
@roe-lz Would it be possible to make a pull request out of your finding so everybody can benefit from it?

Jpaul1001:
sorry, I can't explain the exact function of the entries...
I found them in another forum (NL)...

among other things, there are also other values ​​to read from the Vaillant Arotherm
[...]

PR #599, opened 2026-04-28
vaillant: add RunDataFlowTemp + RunDataReturnTemp to 08.hmu.HW5103

buliwyf42:
## Summary

PR #496 (merged 2025-04-12) added `RunDataFlowTemp` and `RunDataReturnTemp` to `src/vaillant/08.hmu.tsp`. The HW=5103 firmware variant has its own file (`08.hmu.HW5103.tsp`) that does not import from `08.hmu.tsp`, so HW=5103 owners — explicitly the hardware @roe-lz called out in #446 ("These work on `MF=Vaillant;ID=HMU00;SW=0902;HW=5103`") — never received the new entities.

This adds the two models to `08.hmu.HW5103.tsp` so the same telemetry is available on HW=5103 systems.

## Live verification (`HW=5103;SW=0902`)

```
RunDataFlowTemp   = 31.29 °C
RunDataReturnTemp = 31.05 °C
```

Both consistent with the existing `flowtemp` / `status01_temp` + `status01_temp_1` readings within sensor tolerance, but at higher precision because `tempv` is `EXP` rather than `D1C`.

## Build

- `npx tsp compile --emit @ebusd/ebus-typespec --warn-as-error` exits 0
- `ebusd --checkconfig` with the compiled CSV exits 0

## Diff

```diff
   @inherit(r_7)
   @ext(0xde, 0x8)
   model RunDataAirInletTemp {
     value: tempv;
   }

+  @inherit(r_7)
+  @ext(0xfc, 0x8)
+  model RunDataFlowTemp {
+    /** current flow temp, accurate to 2 decimal places */
+    value: tempv;
+  }
+
+  @inherit(r_7)
+  @ext(0x6, 0x9)
+  model RunDataReturnTemp {
+    /** current return temp, accurate to 2 decimal places */
+    value: tempv;
+  }
+
   // B516
```

Same `r_7` base, same `tempv` type, same `@ext` bytes as the generic `08.hmu.tsp`.

### DIFF
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -385,6 +385,20 @@ namespace Hmu_HW5103 {
     value: tempv;
   }
 
+  @inherit(r_7)
+  @ext(0xfc, 0x8)
+  model RunDataFlowTemp {
+    /** current flow temp, accurate to 2 decimal places */
+    value: tempv;
+  }
+
+  @inherit(r_7)
+  @ext(0x6, 0x9)
+  model RunDataReturnTemp {
+    /** current return temp, accurate to 2 decimal places */
+    value: tempv;
+  }
+
   // B516
   /** default *r */
   @base(MF, 0x16)


stgm:
@john30 I'd like these two values to be added, too. They work (tested), and are relevant values on the HW5103. One of these I have been actively using.

Issue #475, opened 2025-01-17
Missing parameters on Arotherm VWZ AI (compr noise reduction, conf.DHW.build.pump)


zarch1972:
Was wondering if anyone has managed to get the following parameters from the VWZ AI heat pump controller available?

- Conf.heat.build.pump
- Conf.cool.build.pump
- Conf.DHW.build.pump
- Compr.noise reduct

It would be great to be able to read (and set these values) remotely rather than having to go to the controller all the time.

Thank you.

zarch1972:
I'm using JonesPD CSV files

https://github.com/jonesPD/ebusd-configuration

```
mick@ebusd:~ $ ebusctl info
version: ebusd 23.3.23.3
update check: version 24.1 available, device firmware 1[4c15] available, broadcast.csv: newer version available, vaillant/08.hmu.csv: different version available, vaillant/broadcast.csv: different version available, vaillant/hcmode.inc: different version available, vaillant/yield3f40.inc: newer version available
device: /dev/ttyAMA0, serial high speed, enhanced, firmware 1.1[3501].1[3501]
signal: acquired
symbol rate: 84
max symbol rate: 410
min arbitration micros: 6
max arbitration micros: 3183
min symbol latency: 0
max symbol latency: 13
scan: finished
reconnects: 0
masters: 6
messages: 870
conditional: 0
poll: 0
update: 18
address 00: master #1
address 03: master #11
address 08: slave #11, scanned "MF=Vaillant;ID=HMU00;SW=0902;HW=5103", loaded "vaillant/08.hmu.csv"
address 10: master #2
address 15: slave #2, scanned "MF=Vaillant;ID=BASV0;SW=0217;HW=8803", loaded "vaillant/15.basv.csv"
address 31: master #8, ebusd
address 36: slave #8, ebusd, scanning
address 71: master #9
address 76: slave #9, scanned "MF=Vaillant;ID=VWZIO;SW=0202;HW=0103", loaded "vaillant/76.vwzio.csv"
address ec: slave, scanned "MF=Vaillant;ID=SOL00;SW=0217;HW=8803", loaded "vaillant/ec.sol.sc.csv"
address f1: master #10
address f6: slave #10, scanned "MF=Vaillant;ID=NETX2;SW=4035;HW=5703"
```

chrizzzp:
@zarch1972 
 
> * Compr.noise reduct

[...]

PR #606, opened 2026-05-07
add SilentModeCompressorReduction for hmu HW5103

maiksensi:
With the recent addition of `SilentTimer_*` I was missing the ability to set the compressor reduction used in silent mode. So here it is.

## Summary
- add `SilentModeCompressorReduction` for `08.hmu.HW5103` using the HMU B51A read/write commands `05ff3428` and `06ff3428`
- expose the field as `percent0` with a 30-60% range

## Validation
- compiled `src/vaillant/08.hmu.HW5103.tsp` with `@ebusd/ebus-typespec` and `withMinMax=true`
- compiled full `src/main.tsp` with `@ebusd/ebus-typespec`
- verified on a live micro-ebusd adapter that the read command returns the configured value and a write of a new value within the range is accepted

## Notes
- this follows the compressor silent operation B51A pattern discussed in #475

### DIFF
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -188,7 +188,7 @@ namespace Hmu_HW5103 {
   @inherit(r_silent, w_silent)
   @ext(0xff, 0x34, 0x28)
   @Ebus.example("40%", "3108b51a0405ff3428", "0aff021b280028003c0001")
-  model SilentModeReductionFactor {
+  model SilentModeCompressorReduction {
     @minValue(30)
     @maxValue(60)
     value: percent0;
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -175,6 +175,25 @@ namespace Hmu_HW5103 {
     value: percent;
   }
 
+  @base(MF, 0x1a, 0x5)
+  model r_silent {
+    @maxLength(3)
+    ign: IGN;
+  }
+
+  @write
+  @base(MF, 0x1a, 0x6)
+  model w_silent {}
+
+  @inherit(r_silent, w_silent)
+  @ext(0xff, 0x34, 0x28)
+  @Ebus.example("40%", "3108b51a0405ff3428", "0aff021b280028003c0001")
+  model SilentModeReductionFactor {
+    @minValue(30)
+    @maxValue(60)
+    value: percent0;
+  }
+
   @inherit(r_2)
   @ext(0x2A)
   model YieldCoolingMonth {


stgm:
@maiksensi I added this setting to #607, too, because my aim is to add all of the settings from the VWZ controller to the config. Did it a bit differently.

Question to you: where did you get the minimum of 30 from? The manuals says 40%.

maiksensi:
The naming is probably debatable. So if there's a better naming let me know and I can refactor that quickly. Something like `SilentModeReductionAmount` is probably better. I don't like adding redundant information like `Comp...` or `percent` that is obvious in that domain. 
Also, awesome that you want to fix all of the settings. I rather like small PRs. Less side effects. But in the end @john30 will probably merge it together anyways :). 
> Question to you: where did you get the minimum of 30 from? The manuals says 40%.

yes, the docs are wrong. At least on my setup I can cycle through 30% - 60%.

stgm:
@maiksensi It's interesting that 30% is possible, given that for the /6 machines 30% is also the absolute minimum for the compressor. Which implies that there would be no modulation if you set this to 30%.

Now that you're here. Do you have an Arotherm Plus /6 machine (not split)? And do you have a system controller / thermostat? Because I'm wondering why this option is hidden in the menus for my setup.

For the names I think something like `Limit` covers even better what it actually does. I'll review the names again later.

stgm:
Oh I now understand it works the other way around (setting 60% here means max 40% speed is still allowed), after explanation at https://github.com/john30/ebusd-configuration/pull/607#discussion_r3217703683

maiksensi:
> Now that you're here. Do you have an Arotherm Plus /6 machine (not split)? And do you have a system controller / thermostat? Because I'm wondering why this option is hidden in the menus for my setup.

It is probably hidden because it is considered an expert setting. So here it is in the expert menu. In German that is "Fachhandwerker" Menu in the VWZ. Yes I do also have a SensoComfort. You can only set the Silent Timer Schedule there.  

> 
> For the names I think something like `Limit` covers even better what it actually does. I'll review the names again later.

I guess, reduction makes even more sense so I updated the PR here.

stgm:
@maiksensi After [some back and forth](https://github.com/john30/ebusd-configuration/pull/607#pullrequestreview-4259704398) I wanted to adopt the terminology from the manual, which is "Noise Reduction" mode. The whole "Silent" is nowhere to be found. So then it became `NoiseReductionLevel`. I dropped Compressor from it, but not 100% sure on that.

stgm:
> It is probably hidden because it is considered an expert setting. So here it is in the expert menu. In German that is "Fachhandwerker" Menu in the VWZ. Yes I do also have a SensoComfort. You can only set the Silent Timer Schedule there.

Sorry, I meant that it's not showing in my expert (installer) menu. Most of the settings appear as in the manual, but this one is simply not there! Just curious, so good to know that you have it.

Issue #535, opened 2025-10-21
Setting / querying DHW pump speed on a VWZ AI


dbrb2:
I have an VWZ AI controlling my ASHP 
I have been tweaking the DHW pump speed from Auto to manual percentage X for a while to try to get around some issues with microbubbles affecting my flow sensor - a setting that is in the Installer configuration section of the VWZ AI

Does anyone know if this is currently tweakable using ebusd? Contorting myself to get to the physical menu under the stairs for every change is a bit tricky!

PR #607, opened 2026-05-10
add base config values from vwzio menu

stgm:
These are most of the settings that can be changed from the VWZ AI install menu on an Arotherm Plus.
Many of these had already been available in the 08.hmu.tsp config, but definitely not all.
Most have been available in the @jonesPD csv repository.

**This PR contains read and write rules for each of the settings.**

- Ready for review, not ready for merging yet.
- I've left out a few values that I do not have (but are mentioned by the manual), I will be taking another look at these.
- Is "max allowed current" too dangerous to expose for writing remotely? Left it out now.

@john30 @burmistrzak @chrizzzp Please do have a look. If some of this is already work in progress, please let me know where to find it.

Should fix at least #475 and #586, #555, #535.

### DIFF
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -379,11 +379,13 @@ namespace Hmu_HW5103 {
     uch: UCH;
   }
 
-  /** maximum anti-cycling time */
+  /** maximum block time by energy company (S21 contact) */
   @inherit(r_3, wi_1)
   @ext(0x34, 0x2d)
-  model MaxAntiCyclingTime {
-    value: integral;
+  model MaximumS21BlockAllowed {
+    @minValue(0)
+    @maxValue(9)
+    value: hour;
   }
 
   /** pump speed for heating */
@@ -392,7 +394,9 @@ namespace Hmu_HW5103 {
   model BuildingCircuitPumpOutputHeating {
     /** pump percentage or auto */
     @values(Values_pumpperc)
-    value: percentp;
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
   }
 
   /** pump speed for cooling */
@@ -401,7 +405,9 @@ namespace Hmu_HW5103 {
   model BuildingCircuitPumpOutputCooling {
     /** pump percentage or auto */
     @values(Values_pumpperc)
-    value: percentp;
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
   }
 
   /** pump speed for domestic hot water */
@@ -410,32 +416,42 @@ namespace Hmu_HW5103 {
   model BuildingCircuitPumpOutputDhw {
     /** pump percentage or auto */
     @values(Values_pumpperc)
-    value: percentp;
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
   }
 
   /** minutes to wait after mains switch on */
   @inherit(r_3, wi_1)
   @ext(0x33, 0x19)
-  model ResetAntiCyclingTime {
+  model BlockTimeAfterRestart {
+    @minValue(0)
+    @maxValue(120)
     value: minutes2;
   }
 
-  /** noise reduction factor applied when in silent mode */
+  /** maximum compressor speed in noise reduction mode */
   @inherit(r_3, wi_1)
   @ext(0x34, 0x28)
   @Ebus.example("40%", "3108b51a0405ff3428", "0aff021b280028003c0001")
-  model CompNoiseReductionFactor {
-    @minValue(30)
+  model CompNoiseReductionSpeed {
+    @minValue(40)
     @maxValue(60)
-    value: percentp;
+    value: percent2;
   }
 
-  /** cooling technology */
+  /** number of main heat pump starts */
+  @inherit(r_3)
+  @ext(0x34, 0x30)
+  model NoOfSwitchingOps {
+    value: cntstarts2;
+  }
+
+  /** cooling enabled */
   @inherit(r_3, wi_1)
   @ext(0x33, 0x1b)
-  model CoolingTechnology {
-    // none / active cooling
-    value: minutes2;
+  model ActiveCoolingEnabled {
+    value: onoff;
   }
 
   // B509
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -228,6 +228,8 @@ namespace Hmu_HW5103 {
     value: energy;
   }
 
+  // Settings that can be changed from the indoors controller unit
+
   /** default *r */
   @base(MF, 0x1a, 0x5, 0xff)
   model r_3 {
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -286,6 +286,7 @@ namespace Hmu_HW5103 {
   model HcAlternativePoint {
     @minValue(-20)
     @maxValue(40)
+    @values(Values_alternative_point)
     temp: temp;
   }
 
@@ -677,5 +678,9 @@ namespace Hmu_HW5103 {
 
   enum Values_pumpperc {
     auto: 49,
+  }
+
+  enum Values_alternative_point {
+    off: -21,
   }
 }
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -231,6 +231,7 @@ namespace Hmu_HW5103 {
   // Settings that can be changed from the indoors controller unit
 
   /** default *r */
+  @auth("install")
   @base(MF, 0x1a, 0x5, 0xff)
   model r_3 {
     @maxLength(3)
@@ -309,14 +310,14 @@ namespace Hmu_HW5103 {
   /** heating function enabled */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x34)
-  model HcModeActive {
+  model HcEnabled {
     yesno: yesno;
   }
 
   /** domestic hot water function enabled */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x35)
-  model HwcModeActive {
+  model HwcEnabled {
     yesno: yesno;
   }
 
@@ -330,7 +331,7 @@ namespace Hmu_HW5103 {
   /** normal immersion heater activation */
   @inherit(r_3, wi_1)
   @ext(0x33, 0x1e)
-  model ImmersionHeaterMode {
+  model ImmersionHeaterEnabled {
     @values(Values_immersion_heater_applications)
     value: UIN;
   }
@@ -338,7 +339,8 @@ namespace Hmu_HW5103 {
   /** emergency immersion heater activation */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x37)
-  model ImmersionHeaterLimpHomeMode {
+  model EmergencyModeImmersionHeaterEnabled {
+    // also called "limp home mode" and "Notbetrieb"
     @values(Values_immersion_heater_applications)
     value: UIN;
   }
@@ -440,7 +442,7 @@ namespace Hmu_HW5103 {
   /** pump speed for domestic hot water */
   @inherit(r_3, wi_1)
   @ext(0x33, 0x18)
-  model BuildingCircuitPumpOutputDhw {
+  model BuildingCircuitPumpOutputHwc {
     /** pump percentage or auto */
     @values(Values_pumpperc)
     @minValue(49)
@@ -461,7 +463,7 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x34, 0x28)
   @Ebus.example("40%", "3108b51a0405ff3428", "0aff021b280028003c0001")
-  model CompNoiseReductionSpeed {
+  model NoiseReductionLevel {
     @minValue(40)
     @maxValue(60)
     value: percent2;
@@ -470,7 +472,7 @@ namespace Hmu_HW5103 {
   /** number of main heat pump starts */
   @inherit(r_3)
   @ext(0x34, 0x30)
-  model HMUStarts {
+  model HMUSystemStarts {
     value: cntstarts2;
   }
 
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -256,6 +256,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x2e)
   model SummerSwitchOffTemp {
+    @minValue(10)
+    @maxValue(90)
     temp: temp;
   }
 
@@ -263,6 +265,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x2f)
   model HcBivalencePoint {
+    @minValue(-30)
+    @maxValue(20)
     temp: temp;
   }
 
@@ -270,6 +274,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x30)
   model HwcBivalencePoint {
+    @minValue(-20)
+    @maxValue(20)
     temp: temp;
   }
 
@@ -277,6 +283,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x31)
   model HcAlternativePoint {
+    @minValue(-20)
+    @maxValue(40)
     temp: temp;
   }
 
@@ -284,6 +292,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x32)
   model MaxFlowTemp {
+    @minValue(15)
+    @maxValue(90)
     temp: temp;
   }
 
@@ -291,17 +301,19 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x33)
   model MinFlowTemp {
+    @minValue(15)
+    @maxValue(90)
     temp: temp;
   }
 
-  /** central heating activation */
+  /** heating function enabled */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x34)
   model HcModeActive {
     yesno: yesno;
   }
 
-  /** domestic hot water activation */
+  /** domestic hot water function enabled */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x35)
   model HwcModeActive {
@@ -312,27 +324,31 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x35, 0x39)
   model HwcChargeHysteresis {
-    calibration: calibration;
+    value: hysteresis;
   }
 
-  /** immersion heater activation */
+  /** normal immersion heater activation */
   @inherit(r_3, wi_1)
   @ext(0x33, 0x1e)
   model ImmersionHeaterMode {
-    onoff: onoff;
+    @values(Values_immersion_heater_applications)
+    value: UIN;
   }
 
-  /** limp home mode activation */
+  /** emergency immersion heater activation */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x37)
-  model LimpHomeMode {
-    //onoff: onoff;
+  model ImmersionHeaterLimpHomeMode {
+    @values(Values_immersion_heater_applications)
+    value: UIN;
   }
 
   /** cooling target flow temperature */
   @inherit(r_3, wi_1)
   @ext(0x35, 0x2b)
-  model CoolTargetFlow {
+  model CoolTargetFlowTemp {
+    @minValue(7)
+    @maxValue(24)
     temp: temp;
   }
 
@@ -340,6 +356,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x34, 0x1e)
   model CompStartHeatingFrom {
+    @minValue(-999)
+    @maxValue(9)
     value: integral;
   }
 
@@ -347,6 +365,8 @@ namespace Hmu_HW5103 {
   @inherit(r_3, wi_1)
   @ext(0x34, 0x21)
   model CompStartCoolingFrom {
+    @minValue(0)
+    @maxValue(999)
     value: integral;
   }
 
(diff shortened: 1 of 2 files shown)


stgm:
I added the percent2 type definition because for the `b51a` 10-byte registers (32-35) with type 02, all data seems to be 2-byte:

`0a f2 0240 3100 3100 6400 01`

- `0a`
    - message length
- `f2`
    - repeat of message identifier from request
- `0240`
    - message type (02/08) and data type
- `3100` <--- this is the real data, always in 2 bytes, always little-endian
    - current value (49)
- `3100`
    - minimum value (49 means auto, 50 is real minimum per the manual)
- `6400`
    - maximum value (100%, conforms to manual)
- `01`
    - divisor

So that's why it seemed appropriate to use a 2-byte data type here. (Also for the new `hysteresis` data type.)

Somewhat related: it is also an option to make a pump speed specific percentage type and add the 49=auto there.

stgm:
I also added the noise reduction setting from #606; putting the format more in line with the rest of the settings in this PR. As it happens, I do not have that setting currently, even though the manual state it should be available (only) if a thermostat/regulator (system control) is *not* connected, like in my situation.

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+  model r_3 {
+    @maxLength(3)
+    ign: IGN;
+  }
+
+  /** default *w */
+  @write
+  @auth("install")
+  @base(MF, 0x1a, 0x6, 0xff)
+  model wi_1 {
```
stgm, line 245: Is `wi` the right abbreviation for write at installer level? (convention-wise)

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+  }
+
+  /** heat curve 0.1 to 4.0 */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x2d)
+  model HeatCurve {
+    @divisor(10)
```
stgm, line 252: Can't provide min/max values 0.4-4.0 for the HeatCurve as the tsp compiler does not allow 0.4 for BCD values. I guess the validation does not take the divisor into account? @john30 

For now I've left the range out and kept it in the description.

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+  model HwcBivalencePoint {
+    @minValue(-20)
+    @maxValue(20)
+    temp: temp;
+  }
+
+  /** alternative point central heating */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x31)
+  model HcAlternativePoint {
```
stgm, line 286: This is new AFAIK, not seen elsewhere.
It's documented in the manual. Pump will leave all heating to the alternative below this temperature (either gas boiler or immersion heater, I think).

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x39)
+  model HwcChargeHysteresis {
+    value: hysteresis;
+  }
+
+  /** normal immersion heater activation */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x1e)
+  model ImmersionHeaterMode {
```
stgm: AFAIK this means whether the immersion heater may be used for heating/water/both/none in normal circumstances (e.g. when heat pump is getting outside working limits?).

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+    @values(Values_immersion_heater_applications)
+    value: UIN;
+  }
+
+  /** emergency immersion heater activation */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x37)
+  model ImmersionHeaterLimpHomeMode {
```
stgm: AFAIK this means whether the immersion heater may be used for heating/water/both/none IF the heat pump fails.

Limp Home mode is from the English manual, the German word "Notbetrieb" is a lot clearer, though more prosaic.

But I can imagine that the gas boiler might also be used for this? It's not clear from the manuals. Anyone??

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+  model CompHysteresisCooling {
+    @minValue(3)
+    @maxValue(15)
+    value: hysteresis;
+  }
+
+  /** maximum allowed differential pressure */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x1f)
+  model MaxRemainingDeltaP {
```
stgm, line 397: This thing has some horribly technical names in the manuals and UI.

> Max. rem. feed head/ differential pressure/ cv max delta P / Restförderhöhe

AFAIK it means the max difference accepted between flow and return temp. I think this concept is well known as the "delta P" so that's what I called it.

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+  @ext(0x34, 0x44)
+  model HwcMode {
+    @values(Values_Domestic_hot_water_mod__eco_normal_balance_)
+    uch: UCH;
+  }
+
+  /** maximum block time by energy company (S21 contact) */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x2d)
+  model MaximumS21BlockAllowed {
```
stgm, line 415: This was fairly lost in manual translations:

- max. Dauer Sperrzeit
- Max. **anti-cycl**. time 

The translation suggests it has to do with cycling (pump on/off frequently?). 
Digging through the manuals makes clear that Sperrzeit is linked to the power company blocking contact S21. So I named it that.

@jonesPD has MainsBlocktimeMaxDuration

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+    value: hour;
+  }
+
+  /** pump speed for heating */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x16)
+  model BuildingCircuitPumpOutputHeating {
+    /** pump percentage or auto */
+    @values(Values_pumpperc)
+    @minValue(49)
```
stgm, line 427: 49 as "auto" experimentally verified. Register response also makes clear that 49-100 are valid values.

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+    @maxValue(100)
+    value: percent2;
+  }
+
+  /** pump speed for domestic hot water */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x18)
+  model BuildingCircuitPumpOutputDhw {
```
stgm: Probably should make this Hwc instead of Dhw

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+    @values(Values_pumpperc)
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
+  }
+
+  /** minutes to wait after mains switch on */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x19)
+  model BlockTimeAfterRestart {
```
stgm, line 457: Took some digging, in the manual this is:

- Reset Sperrzeit→ Sperrz. n. Einschalten der Spannungsvers.
- Reset anti-cycl. time→ Anti-cycl. time after pow. supp. switch-on

Sperrzeit is again translated as "Anti-cycling". But focusing on "after pow. supp. switch-on" and the German description makes it pretty clear what is going on; in combination with time in minutes and not hours, this seems what it is.

@jonesPD has `MainsBlocktimeMinDuration` here which I think is a mistake after all. It does not seem to have anything to do with the `MainsBlocktimeMaxDuration` or what I now call `MaximumS21BlockAllowed`

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+    @minValue(0)
+    @maxValue(120)
+    value: minutes2;
+  }
+
+  /** maximum compressor speed in noise reduction mode */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x28)
+  @Ebus.example("40%", "3108b51a0405ff3428", "0aff021b280028003c0001")
+  model CompNoiseReductionSpeed {
```
stgm: I do not have this one! 

Do you have it @chrizzzp?

stgm: #606 has actually proposed this. 

I've tried to come to a name that more clearly describes the variable.

@jonesPD has `CompressorSilentOperationLevel` which I think is pretty good, though it really should be interpreted as a maximum, AFAIK.

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
...
+    @minValue(40)
+    @maxValue(60)
+    value: percent2;
+  }
+
+  /** number of main heat pump starts */
+  @inherit(r_3)
+  @ext(0x34, 0x30)
+  model HMUStarts {
+    value: cntstarts2;
```
stgm, line 477: Otherwise known as `NoOfSwitchingOps` it was one of the few things that is not in the vwzio config menus but also missing in the config file. It's in the jonesPD repo.

Turning my heat pump off and then on increased the counter, so I _guess_ `HMUStarts` is a good name.

Still reflecting on whether this one belongs in the PR, because it's not available from the menus.

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,259 @@ namespace Hmu_HW5103 {
     value: energy;
   }
 
+  // Settings that can be changed from the indoors controller unit
+
+  /** default *r */
+  @base(MF, 0x1a, 0x5, 0xff)
+  model r_3 {
```
stgm, line 236: @john30 Should read rules also be "installer" level?

I keep hearing about polling (maybe that's a home assistant thing, which I don't use).. these values obviously do not need to be polled. But reading them is safe, so normally I think no extra auth would be needed.

But maybe you invented the install level to prevent polling problems? I can't find about the why and how of these authentication levels.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: There is (maybe) similar setting if a regulator is used:

from [this PR](https://github.com/john30/ebusd-configuration/pull/564)
15.720.csv:
```
r,,install,BackupBoiler,backup boiler,,,b524,020000002400,ign,,IGN:4,,,,value,,UCH,0=off;1=heat;2=water;3=heatandwater,,
w,,install,BackupBoiler,backup boiler,,,b524,020100002400,value,,UCH,0=off;1=heat;2=water;3=heatandwater,,
```

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: I think it's a difference pressure, so unit should be bar/(mbar with diveder 1000) and affects building circuit pump power.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: Correct, it is the maximum time S21 can be blocked (by the power company) before the HMU will throw an error.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: On my split system this parameter is unavailable (operated with a regulator).

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: I agree. For my system the value is 0. Never tried it out.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: Yes, it works. For my split system it can be adjust between 40 and 60%.

Actually, a value of 60% will limit the HMU power to 40% (and vice versa). So it's rather a throttle or reduction  value, not a direct limit. 

So how about: `CompressorSilentModeReduction` or `CompressorSilentModeThrottle`?

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: Not to mix with `hmu CompressorStarts`. So maybe `HMUSystemStarts`?

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: IMO Polling = a master sending a read command via Ebus (directly or by MQTT) and ebusd decoding the slave reply

I think the elevated auth levels for write commands make sense to (sort of) prevent possible misconfiguration of the system or important parameters. For read commands they could be used to hide certain system settings from (standard) users (similar to the settings behind the 'Installer level' menus of the VWZ UI or regulator UI. But it is easily overridden by the --acccesslevel=* option of ebusd, so the level of protection is limited.
However, I think we should implement auth levels and stick to the Vaillant scheme of the 'Installer Level' options.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: Yields `02ff01` (not applicable) for my system (with regulator). I'm not aware of any similar option in the regulator, unless it's an (emergency) temperature definition:

```
r,,install,HcEmergencyTemp,hc emergency temperature,,,b524,020000002600,ign,,IGN:4,,,,value,,EXP,,°C,temperature
w,,install,HcEmergencyTemp,hc emergency temperature,,,b524,020100002600,value,,EXP,,°C,temperature
```

src/vaillant/08.hmu.HW5103.tsp
stgm: @chrizzzp Hmm I assumed that the pump would actually _ignore_ the block after the set time (as in: it would start heating again), but you say it will error?

src/vaillant/08.hmu.HW5103.tsp
stgm: In my case (not split) the system actually showed me the setting (in the VWZ AI controller), but did not effectively change it to another value when it chose that. So that behaviour is a bit weird.

I will do some archival research (aka old manuals) to see how this developed. At least I already found somewhere that "limp mode" comes from gas boilers that might continue working on lower modulation until the repair-person comes. So they may have transferred an old term to the new context of heat pumps at some point.

And now that you say it, I'll also take the regulator manuals to see if those have something relevant.

src/vaillant/08.hmu.HW5103.tsp
stgm: @chrizzzp To be clear: `BuildingCircuitPumpOutputHeating` does work? And you have a three-way valve somewhere? Don't think it matters much, but still curious.

stgm:
Thank you for all the feedback @chrizzzp !

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: IMO the S21 block is not to be ignored by the heating system. 
But to notice the owner of the system of a prolonged block (or power cut, see below) this time limit can be defined.

From my experience, when I cut the 3-phase power of the HMU (400V wiring) the UI of the hydraulic station is still powered by the 230V connection and will report after `MainsBlocktimeMaxDuration` hours an error (3-phase power outtage).

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: No, `BuildingCircuitPumpOutputHeating` not working.
`08b51a0405ff3316 / 02ff01`

But I have a related register in the `b514` block:
`hmu BuildingCircuitPumpPower = 3108b51405050103ffff / 0401003200`

stgm: Ah yes I see. They are indeed missing in the manual for the split.

src/vaillant/08.hmu.HW5103.tsp
stgm: The read rules are also installer-level now.

src/vaillant/08.hmu.HW5103.tsp
stgm: Confirmed from the description in the vrc720f/2 manual.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: Forgot to mention: yes, of course,  there is a three-way-valve also in the split.

src/vaillant/08.hmu.HW5103.tsp
stgm: A sliver of info from latest German manual:

> Der Systemregler schaltet in den Notbetrieb um, d. h. die
Elektro-Zusatzheizung versorgt die Heizungsanlage mit Hei-
zenergie. Der Fachhandwerker hat bei der Installation für
den Notbetrieb die Temperatur gedrosselt. Sie spüren, dass
das Warmwasser und die Heizung nicht sehr warm werden.

In other words: if the pump fails, this setting decides whether the immersion heating element may be used as an alternative.

-> Or actually, whatever is attached to the ZH (zusatzheizung) contact on the vwz AI. It is also used to attach non-Vaillant gas boilers in hybrid systems.

The explanation also mentions setting a lower target temperature, to help you notice that the heating is not working right. Which is apparently what `HcEmergencyTemp` is for. But only available if you have a regulator?!

All in all, a better name for this one is probably "EmergencyModeUseImmersionHeater" (meh).

src/vaillant/08.hmu.HW5103.tsp
stgm: Unclear if it is actually the same - it seems that the **hmu** setting is specifically for anything that's connected to the ZH contact in the Vwz AI (see below).

But the **regulator** seems to have additional stuff to enable hybrid setups. So BackupBoiler may refer to the hybrid gas boiler specifically. In the Dutch manual this setting specifically refers to the gas boiler, not the immersion heater.

Still I'd rename this to `ImmersionHeaterEnabled` then.

src/vaillant/08.hmu.HW5103.tsp
stgm: Interesting :-)
Should I keep the jonesPD name? I found it slightly confusing (when I knew less about it).

src/vaillant/08.hmu.HW5103.tsp
stgm: Any reason you're converging to "silent mode" @chrizzzp ?
The English manual for the plus/6 specifically calls it "noise reduction mode". (Oh and the device menu, too.)

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: I agree with your nomenclature. The german manual ("Max. Dauer Sperrzeit") indeed suggests, the blocking is ignored after this period. I think this will need experimental proof...

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: So this is rather a startup delay value, isn't it? Lacks experimental proove I assume.

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: It's shorter. :) 
But you're right we should leave it to the (slightly awkward) nomenclature by Vaillant. However, then we also should rename the 'Silent Timer' to 'Noise Reduction Timer' in the [regulator PR](https://github.com/john30/ebusd-configuration/pull/564) for the sake of consistency.

BTW: Do you know [this](https://simulator.vaillant.com/vrc720_3/hu/#/setup) regulator simulator?

stgm: Yes, I have used it a few times!

I think "CompNoiseReductionLevel" then. Considering how it works, level seems a good concept. Still thinking about the "Comp" prefix which is very non-informative.

src/vaillant/08.hmu.HW5103.tsp
stgm: Yes and yes.

src/vaillant/08.hmu.HW5103.tsp
stgm: Yeah this is one I can't do myself ;-)

src/vaillant/08.hmu.HW5103.tsp
chrizzzp: Maybe we can "afford" `ComprNoiseReductionLevel` or even `CompressorNoiseReductionLevel`?

src/vaillant/08.hmu.HW5103.tsp
stgm: OK `EmergencyModeImmersionHeaterEnabled`.
I'm deviating from official "limp home" wording here, because it's very tongue-in-cheek British and unlikely that people recognize the word, especially from other countries.
I'd love to do something with a more literal version of "zusatzheizung" but it gets long anyway.

src/vaillant/08.hmu.HW5103.tsp
stgm: Oh missed your message before I typed mine (but now changed).
I think this variable is not necessarily about the compressor per se (although that's what the percentage is applied to)
NoiseReductionLevel could suffice.
This in contrast to CompStartHeatingFrom which really is about the compressor algorithm.

src/vaillant/08.hmu.HW5103.tsp
stgm: To be clear: I think my suggestion already completely satisfies your remark.

pressm2 -> mbar

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,261 @@ namespace Hmu_HW5103 {
...
+  model MinFlowTemp {
+    @minValue(15)
+    @maxValue(90)
+    temp: temp;
+  }
+
+  /** heating function enabled */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x34)
+  model HcEnabled {
```
stgm, line 314: Use of "enabled" instead of "active" - in line with
https://github.com/john30/ebusd-configuration/pull/564/changes

Same for `HwcEnabled` below

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -228,6 +228,261 @@ namespace Hmu_HW5103 {
...
+    yesno: yesno;
+  }
+
+  /** domestic hot water function enabled */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x35)
+  model HwcEnabled {
```
stgm, line 321: I think @chrizzzp and @burmistrzak had a discussion somewhere about Hwc vs Dhw, right? Hwc it is?

src/vaillant/08.hmu.HW5103.tsp
```diff
@@ -406,4 +662,25 @@ namespace Hmu_HW5103 {
...
+    heating: 1,
+    dhw: 2,
+    both: 3,
+  }
+
+  enum Values_pumpperc {
+    auto: 49,
+  }
+
+  enum Values_alternative_point {
```
stgm, line 683: @john30 There are quite a few variables with a single `off` or `auto` value. Is there a shorter way to express this other than defining an enum? I see the value of the enums for multiple values, but for a single value it's getting a bit messy.

PR #611, opened 2026-05-27
[codex] vaillant: generate ctls3 from 700 and add cooling setpoints

Damecek:
## What changed

This adds a `15.ctls3` alias to the 700 controller definition and fills in the missing zone cooling setpoint registers in `15.700.tsp`.

Specifically:
- add `src/vaillant/15.ctls3.tsp` as a symlink to `15.700.tsp`
- add `Z1/Z2/Z3CoolingTempDesired` on offset `0x15`
- add `Z1/Z2/Z3CoolingManualTemp` on offset `0x23`

## Why

A Vaillant Genia Air split installation with a scanned `CTLS3` controller was using `vaillant/15.ctls3.csv`, but the active cooling target changed in the UI without any matching attribute in the available `ctls3` definitions.

Runtime verification on the target system showed that the controller exposes:
- `Z2CoolingTempDesired` at `020003011500`
- `Z2CoolingManualTemp` at `020003012300`

Those attributes were missing from the generated `ctls3` config, while the rest of the observed CTLS3 layout matched the 700-style room controller model:
- `Z2RoomZoneMapping = VRC700`
- `Z2OpMode = off/auto/day/night`
- `Hc1CircuitType = mixer`

## User impact

After publishing this, a generated `vaillant/15.ctls3.csv` can expose the cooling setpoints that are actually used by CTLS3-backed systems.

Expected MQTT topics include:
- `GeniaAir-split/ctls3/Z2CoolingTempDesired`
- `GeniaAir-split/ctls3/Z2CoolingManualTemp`

## Validation

Checks run:
- `npm run compile-en`
- `npm run lint`

Runtime verification on the real CTLS3 setup was done with `ebusctl` against a temporary `ebusd` instance using the updated generated config:
- `Z2CoolingTempDesired = 18.5`
- `Z2CoolingManualTemp = 18.5`
- `Z2CoolingTemp = 19`
- `Z2RoomZoneMapping = VRC700`

This confirms that the newly defined attributes map to the active cooling target, while the legacy `Z2CoolingTemp` remains a different setpoint.


Narcotic82:
I tested these six new CTLS3 cooling registers on another real CTLS3 SW0809 / HW8104 installation using ebusd 26.1.26.1.

All six registers are recognized and return valid values:

Z1CoolingTempDesired = 27
Z1CoolingManualTemp = 27

Z2CoolingTempDesired = 0.0
Z2CoolingManualTemp = 24

Z3CoolingTempDesired = 0.0
Z3CoolingManualTemp = 24

The definitions therefore work correctly on my CTLS3 SW0809/HW8104 installation as well.

This provides an additional live validation of the proposed CTLS3 cooling register definitions.

PR #622, opened 2026-06-24
vaillant: document recoVair 275/3 findings (VAI00 SW=0125, WTW04 SW=0127 HW=9001)

HerrBausW:
## Summary

Validated eBUS register behavior with a **Vaillant recoVair 275/3** ventilation unit:
- VAI00 SW=0125 HW=8901
- WTW04 SW=0127 HW=9001

### 15.vai.tsp

- **Mode register (0x24)**: Documented actual values for the recoVair: `on(1)=day`, `off(2)=off`, `auto(3)=auto`, `low(5)=night`.
- **HighDuration register (0x58)**: Uses **HCD:2 encoding** (not plain UIN minutes). Values are 0.5h steps: 0=off, 2=0.5h, 3=1h, ..., 7=3h. The device decrements the value by one step immediately on activation.

### c0.wtw.tsp

- **Temperature registers on HW=9001**: Require **PBSB B509,29 with a 2-byte IGN prefix** instead of the standard B509,0D. Without this, temperature reads return invalid data on this hardware revision.
- Clarified temperature sensor names (Zuluft/Außenluft/Abluft/Fortluft).

## Test setup

- ebusd 26.1 on Debian 12
- eBUS TCP adapter
- Full read/write validation of all documented registers


HerrBausW:
Closing this PR due to incorrect findings about HighDuration. After further testing: the register counts in MINUTES (1 unit = 1 minute), not in 0.5h steps. The device decrements by 1 immediately upon activation, then by 1 per minute thereafter. The original minutes2 type may therefore be more appropriate than the HCD2 change introduced here. Additionally, the temperature PBSB note for WTW04 HW=9001 (B509,29 with IGN:2 prefix) remains valid and may be contributed separately once better understood in TypeSpec.

PR #649, opened 2026-08-27
Update 08.hmu.HW5103.tsp

ClausLDK:
Updated Unit correct to L/h (BuildingCircuitFlow)


twslocombe:
This is incorrect. It should be    

@unit("l/hour")

ClausLDK (commented):
corrected to l/hour

ClausLDK:
> This is incorrect. It should be
> 
> @Unit("l/hour")

No, according to latest HA version it should be L/h

twslocombe:
Both "L/h" and "l/hour" work. If "L/h" is in the official documentation, this is the correct one to use

PR #652, opened 2026-08-28
Update 08.hmu.HW5103.tsp

ClausLDK:
BuildingCircuitFlow update to correct unit l/h -> L/h


ClausLDK:
Consolidated in https://github.com/john30/ebusd-configuration/pull/660

Issue #570, opened 2026-02-17
myVAILLANT: unknown MS cmd: tips


Commifreak:
Hey,

I have a VR921 with ambiSense.

I observed all values which are getting changed, when an eTRV doesnt need any heat and instantly saw, that the ZxOpMode and daytemp gets changed from address `f1`:

```
address f1: master #10
address f6: slave #10, scanned "MF=Vaillant;ID=NETX2;SW=4038;HW=5703"
```

Nice. Next I wanted to see what the gateway issues, when I want a Hwc load instantly - and saw NOTHING on the bus? but CirPump and HwcSFMode getting changed somehow - but I see those changes just over the poll-read. Theres nothing that wrote it?

Interesting. But I remembered many `received unknown MS cmd:` log lines and started to enabled/disable hwcLoading via App.

I think `f115b52407020101ff0d0006 / 020001` means: enable and `f115b52407020101ff0d0000 / 020001` disable. I tried it several times and those two are there at any try - along with others.

Are there any informations for the messages coming from `f1`?

I have several from these. If I can help improving config/templates, Iam happy to help :)

```
2026-02-17 18:09:31.801 [update notice] received unknown MS cmd: f115b52406020003011b00 / 0601031b000000
2026-02-17 18:09:32.006 [update notice] received unknown MS cmd: f115b52406020001000f00 / 0601010f000900
2026-02-17 18:09:47.390 [update notice] received unknown MS cmd: f115b52406020003001b00 / 0601031b000000
2026-02-17 18:09:47.596 [update notice] received unknown MS cmd: f115b52406020003011b00 / 0601031b000000
2026-02-17 18:09:47.801 [update notice] received unknown MS cmd: f115b52406020001000f00 / 0601010f000900
2026-02-17 18:09:59.880 [update notice] received unknown MS cmd: 1008b5110100 / 08a20411be183f0080
2026-02-17 18:10:03.394 [update notice] received unknown MS cmd: f115b52406020003001b00 / 0601031b000000
2026-02-17 18:10:03.601 [update notice] received unknown MS cmd: f115b52406020003011b00 / 0601031b000000
2026-02-17 18:10:03.803 [update notice] received unknown MS cmd: f115b52406020001000f00 / 0601010f000900
2026-02-17 18:10:18.889 [update notice] received unknown MS cmd: f115b503020002 / 0affffffffffffffffffff
[...]

PR #659, opened 2026-09-04
vaillant: add SystemDemand for CTLV2

ClausLDK:
Adds decoding for B524 register 0x48 as SystemDemand on CTLV2/BASV3. Observed values on real hardware: 0=off, 1=heating, 3=hot_water. 2=cooling is currently assumed. Related to #570.


ClausLDK:
@john30 - tested via local and verified, can you please add?

PR #614, opened 2026-06-05
vaillant: add Status07 (display/operating mode) to 08.hmu.HW5103

buliwyf42:
## Summary

Adds `Status07` (`b511 07`) to `08.hmu.HW5103.tsp`, restoring the heat-pump operating mode ("Betriebsart") for HW=5103 owners. **Closes #609.**

The legacy generic `08.hmu.csv` exposed the operating mode (heating / cooling / hot water / error) as field 4 of the old "State" message on `b511 07`. The HW=5103 variant file never carried `b511 07`, so that value disappeared for HW=5103 users.

As diagnosed in #609, there is no working substitute among the other `b511` messages: on this hardware **`Status02` (`b511 02`) and `Status` (`b511 03`) return an empty body to an active read**, which is exactly the `ERR: invalid position in decode` the reporter hit. Reproduced on my own HW=5103:

```
hmu Status02 =  (ERR: invalid position for 3108b5110102 / 00)
hmu Status   =  (ERR: invalid position for 3108b5110103 / 00)
```

## What it decodes

`Status07` is a 7-byte poll message:

| field | type | meaning |
|---|---|---|
| `power` | UCH % | compressor power |
| `dailyenvyield` | UIN /10 kWh | daily environmental energy yield |
| `display_b0..b7` | bit flags | b0=heater enabled, b2=backup heater active, b5=noise reduction, b6=DHW eco mode (per @chrizzzp in #598); b1/b3/b4/b7 reserved |
| `heatermain_b0..b7` | bit flags | **heater main operating mode** (see below) |
| `DisplayPressure` | UCH /30 bar | system pressure |
| `heaterbackup_b0..b7` | bit flags | backup heater operating mode (same encoding) |

## Operating-mode byte is a bit field

Thanks to @hanjo (the #609 reporter, who uses this value in production to drive his underfloor heating and to detect "Komfortsicherung"), the mode byte is modelled as a **bit field** rather than an enum:

| bit | meaning |
|---|---|
| b1 | error active (Komfortsicherung) |
| b3 | heating |
| b4 | cooling |
| b5 | pressure loss |
| b7 | hot water |
| b0 / b2 / b6 | reserved (not yet observed) |

[...]

Issue #638, opened 2026-07-23
PowerConsumptionHmu and Errorhistory decode errors in 08.hmu.HW5103.csv (Vaillant aroTHERM)


MarkBovee:
## Device

**Hardware:** Vaillant aroTHERM VWL 55/5 AS  
**Address 08 (HMU):** MF=Vaillant, ID=HMU00, SW=0522, HW=5103  
**ebusd version:** 26.1.26.1  
**Config source:** `https://ebus.github.io/de/`  
**Loaded file:** `vaillant/08.hmu.HW5103.csv`

## Bug 1: PowerConsumptionHmu (b516,14) — "invalid position"

**CSV line:** `r,,,PowerConsumptionHmu,,,,b516,14,ign,,IGN:1,,,,value,,EXP,1000,kW,`

**Raw hex query:** `3108b5160114` (QQ=31, ZZ=08, PB=B5, SB=16, NN=01, DD=14)  
**Raw response:** `00` (single byte 0x00)

**Problem:** The definition has `IGN:1` (skip 1 byte) followed by `EXP` (2+ byte exponential). The bus response is only **1 byte total**. The `IGN:1` consumes that byte, then the `EXP` decode at position 1 fails with "invalid position" because there is no second byte.

The `00` response likely **is** the power value (0 W when compressor is idle), making the `IGN:1` incorrect — it should not be there. If we strip the IGN, the value should decode as a single-byte unsigned integer with divisor 1000 → 0 kW.

## Bug 2: Errorhistory (b503,0101) — "invalid position"

**CSV line:** `r,,,Errorhistory,Fehlerhistorie,,,b503,0101,index,m,UCH,,,,status,,UCH,,,Status,time,,VTM,,,Uhrzeit,date,,HDA:3,,,Datum,error,,UIN,,,Fehlernummer`

**Raw hex query:** `3108b503020101` (QQ=31, ZZ=08, PB=B5, SB=03, NN=02, DD=0101)  
**Raw response:** `00` (single byte 0x00)

**Problem:** The definition expects 5+ fields (index UCH, status UCH, time VTM, date HDA:3, error UIN). The response is only **1 byte**. The first field `index(UCH)` reads byte 0x00 successfully, but `status(UCH)` at position 1 fails.

`00` likely means "no error at this index" or "error list is empty" (hwc compressor not running, no errors to report).

## Notes

[...]

PR #660, opened 2026-09-04
vaillant: consolidate HW5103 configuration updates

ClausLDK:
Consolidates several pending HW5103 configuration improvements into one tested configuration. Includes the base/install configuration values and required percent2/hysteresis types from #607 (including the noise-reduction work related to #606), Status07 operating-mode decoding from #614. Tested on a real Vaillant HMU HW5103 system and compiled successfully with TypeSpec 1.12.0. PowerConsumptionHmu is already present in current master and is not modified here. Credit to @stgm, @buliwyf42, @maiksensi and the contributors/reporters on the referenced PRs and issues.

### DIFF
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -1,7 +1,6 @@
 import "@ebusd/ebus-typespec";
 import "./_templates.tsp";
 import "./hcmode_inc.tsp";
-import "./errors_inc.tsp";
 using Ebus;
 using Ebus.Num;
 using Ebus.Dtm;
@@ -30,6 +29,125 @@ namespace Hmu_HW5103 {
     @maxLength(1) ign: IGN;
     runtime: minutes4;
     cycles: cntstarts;
+  }
+
+/** HMU display state: compressor power, environmental yield and heater main/backup operating mode */
+  @inherit(r_1)
+  @ext(0x7)
+  model Status07 {
+    /** compressor power */
+    @unit("%")
+    power: UCH;
+
+    /** daily environmental energy yield */
+    @divisor(10)
+    dailyenvyield: energy;
+
+    /** heater main enabled */
+    @values(Values_onoff)
+    display_b0_heaterenabled: BI0;
+
+    /** reserved bit 1 */
+    @values(Values_onoff)
+    display_b1: BI1;
+
+    /** backup heater active */
+    @values(Values_onoff)
+    display_b2_backupheater: BI2;
+
+    /** reserved bit 3 */
+    @values(Values_onoff)
+    display_b3: BI3;
+
+    /** reserved bit 4 */
+    @values(Values_onoff)
+    display_b4: BI4;
+
+    /** noise reduction active */
+    @values(Values_onoff)
+    display_b5_noisereduction: BI5;
+
+    /** DHW eco mode */
+    @values(Values_onoff)
+    display_b6_dhwecomode: BI6;
+
+    /** reserved bit 7 */
+    @values(Values_onoff)
+    display_b7: BI7;
+
+    // heater main operating mode (bit field; bit semantics per @hanjo in #614,
+    // confirmed against a 72 h bus capture: only standby/heating/warmwater seen here)
+
+    /** reserved bit 0 */
+    @values(Values_onoff)
+    heatermain_b0: BI0;
+
+    /** error active (Komfortsicherung) */
+    @values(Values_onoff)
+    heatermain_b1_error: BI1;
+
+    /** reserved bit 2 */
+    @values(Values_onoff)
+    heatermain_b2: BI2;
+
+    /** heating active */
+    @values(Values_onoff)
+    heatermain_b3_heating: BI3;
+
+    /** cooling active */
+    @values(Values_onoff)
+    heatermain_b4_cooling: BI4;
+
+    /** pressure loss */
+    @values(Values_onoff)
+    heatermain_b5_pressureloss: BI5;
+
+    /** reserved bit 6 */
+    @values(Values_onoff)
+    heatermain_b6: BI6;
+
+    /** hot water active */
+    @values(Values_onoff)
+    heatermain_b7_warmwater: BI7;
+
+    /** system pressure */
+    @unit("bar")
+    @divisor(30)
+    DisplayPressure: UCH;
+
+    // backup heater operating mode (same bit field as the main heater)
+
+    /** reserved bit 0 */
+    @values(Values_onoff)
+    heaterbackup_b0: BI0;
+
+    /** error active (Komfortsicherung) */
+    @values(Values_onoff)
+    heaterbackup_b1_error: BI1;
+
+    /** reserved bit 2 */
+    @values(Values_onoff)
+    heaterbackup_b2: BI2;
+
+    /** heating active */
+    @values(Values_onoff)
+    heaterbackup_b3_heating: BI3;
+
+    /** cooling active */
+    @values(Values_onoff)
+    heaterbackup_b4_cooling: BI4;
+
+    /** pressure loss */
+    @values(Values_onoff)
+    heaterbackup_b5_pressureloss: BI5;
+
+    /** reserved bit 6 */
+    @values(Values_onoff)
+    heaterbackup_b6: BI6;
+
+    /** hot water active */
+    @values(Values_onoff)
+    heaterbackup_b7_warmwater: BI7;
   }
 
   /** default *r */
@@ -208,7 +326,7 @@ namespace Hmu_HW5103 {
   @inherit(r_2)
   @ext(0x3c)
   model BuildingCircuitFlow {
-    @unit("l/h")
+    @unit("L/h")
     value: UIN;
   }
 
@@ -226,6 +344,259 @@ namespace Hmu_HW5103 {
   model TotalEnergyUsage {
     @unit("kWh")
     value: energy;
+  }
+
+  // Settings that can be changed from the indoors controller unit
+
+  /** default *r */
+  @auth("install")
+  @base(MF, 0x1a, 0x5, 0xff)
+  model r_3 {
+    @maxLength(3)
+    ign: IGN;
+  }
+
+  /** default *w */
+  @write
+  @auth("install")
+  @base(MF, 0x1a, 0x6, 0xff)
+  model wi_1 {
+  }
+
+  /** heat curve 0.1 to 4.0 */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x2d)
+  model HeatCurve {
+    @divisor(10)
+    bcd: BCD;
+  }
+
+  /** summer switch-off temperature */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x2e)
+  model SummerSwitchOffTemp {
+    @minValue(10)
+    @maxValue(90)
+    temp: temp;
+  }
+
+  /** bivalence point central heating */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x2f)
+  model HcBivalencePoint {
+    @minValue(-30)
+    @maxValue(20)
+    temp: temp;
+  }
+
+  /** bivalence point domestic hot water */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x30)
+  model HwcBivalencePoint {
+    @minValue(-20)
+    @maxValue(20)
+    temp: temp;
+  }
+
+  /** alternative point central heating */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x31)
+  model HcAlternativePoint {
+    temp: temp;
+  }
+
+  /** maximum allowed flow temperature */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x32)
+  model MaxFlowTemp {
+    @minValue(15)
+    @maxValue(90)
+    temp: temp;
+  }
+
+  /** minimum allowed flow temperature */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x33)
+  model MinFlowTemp {
+    @minValue(15)
+    @maxValue(90)
+    temp: temp;
+  }
+
+  /** heating function enabled */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x34)
+  model HcEnabled {
+    yesno: yesno;
+  }
+
+  /** domestic hot water function enabled */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x35)
+  model HwcEnabled {
+    yesno: yesno;
+  }
+
+  /** hysteresis for cylinder charging */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x39)
+  model HwcChargeHysteresis {
+    value: hysteresis;
+  }
+
+  /** normal immersion heater activation */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x1e)
+  model ImmersionHeaterEnabled {
+    @values(Values_immersion_heater_applications)
+    value: UIN;
+  }
+
+  /** emergency immersion heater activation */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x37)
+  model EmergencyModeImmersionHeaterEnabled {
+    // also called "limp home mode" and "Notbetrieb"
+    @values(Values_immersion_heater_applications)
+    value: UIN;
+  }
+
+  /** cooling target flow temperature */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x2b)
+  model CoolTargetFlowTemp {
+    @minValue(7)
+    @maxValue(24)
+    temp: temp;
+  }
+
+  /** start heating at energy integral value */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x1e)
+  model CompStartHeatingFrom {
+    @minValue(-999)
+    @maxValue(9)
+    value: integral;
+  }
+
+  /** start cooling at energy integral value */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x21)
+  model CompStartCoolingFrom {
+    @minValue(0)
+    @maxValue(999)
+    value: integral;
+  }
+
+  /** hysteresis for heating */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x43)
+  model CompHysteresisHeating {
+    @minValue(3)
+    @maxValue(15)
+    value: hysteresis;
+  }
+
+  /** hysteresis for cooling */
+  @inherit(r_3, wi_1)
+  @ext(0x35, 0x3f)
+  model CompHysteresisCooling {
+    @minValue(3)
+    @maxValue(15)
+    value: hysteresis;
+  }
+
+  /** maximum allowed differential pressure */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x1f)
+  model MaxRemainingDeltaP {
+    @minValue(200)
+    @maxValue(900)
+    @step(10)
+    value: pressm2;
+  }
+
+  /** domestic hot water charging mode */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x44)
+  model HwcMode {
+    @values(Values_Domestic_hot_water_mod__eco_normal_balance_)
+    uch: UCH;
+  }
+
+  /** maximum block time by energy company (S21 contact) */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x2d)
+  model MaximumS21BlockAllowed {
+    @minValue(0)
+    @maxValue(9)
+    value: hour;
+  }
+
+  /** pump speed for heating */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x16)
+  model BuildingCircuitPumpOutputHeating {
+    /** pump percentage or auto */
+    @values(Values_pumpperc)
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
+  }
+
+  /** pump speed for cooling */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x17)
+  model BuildingCircuitPumpOutputCooling {
+    /** pump percentage or auto */
+    @values(Values_pumpperc)
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
+  }
+
+  /** pump speed for domestic hot water */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x18)
+  model BuildingCircuitPumpOutputHwc {
+    /** pump percentage or auto */
+    @values(Values_pumpperc)
+    @minValue(49)
+    @maxValue(100)
+    value: percent2;
+  }
+
+  /** minutes to wait after mains switch on */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x19)
+  model BlockTimeAfterRestart {
+    @minValue(0)
+    @maxValue(120)
+    value: minutes2;
+  }
+
+  /** maximum compressor speed in noise reduction mode */
+  @inherit(r_3, wi_1)
+  @ext(0x34, 0x28)
+  @Ebus.example("40%", "3108b51a0405ff3428", "0aff021b280028003c0001")
+  model NoiseReductionLevel {
+    @minValue(40)
+    @maxValue(60)
+    value: percent2;
+  }
+
+  /** number of main heat pump starts */
+  @inherit(r_3)
+  @ext(0x34, 0x30)
+  model HMUSystemStarts {
+    value: cntstarts2;
+  }
+
+  /** cooling enabled */
+  @inherit(r_3, wi_1)
+  @ext(0x33, 0x1b)
+  model ActiveCoolingEnabled {
+    value: onoff;
   }
 
   // B509
@@ -385,6 +756,40 @@ namespace Hmu_HW5103 {
     value: tempv;
   }
 
+  // Error handling for this HW5103 variant.
+  //
+  // Errorhistory (b503 0101) is intentionally not exposed here. On this HMU
+  // the request returns a single byte 0x00 when no history entry is available,
+  // while the generic Errors_inc.Errorhistory definition expects a full
+  // status/time/date/error record and therefore fails with "invalid position".
+  // See ebusd-configuration issue #638.
+  // Currenterror and Clearerrorhistory are kept unchanged.
+
+  /** default *r for current errors */
+  @base(MF, 0x3, 0)
+  model r_error_current {}
+
+  /** current errors */
+  @inherit(r_error_current)
+  @ext(1)
+  model Currenterror {
+    value: errors;
+  }
+
+  /** default *wi for clearing error history */
+  @write
+  @auth("install")
+  @base(MF, 0x3, 0x2)
+  model wi_error_clear {}
+
+  /** clear error history */
+  @inherit(wi_error_clear)
+  @ext(1)
+  model Clearerrorhistory {
+    @in
+    cleared: yesno;
+  }
+
   // B516
   /** default *r */
   @base(MF, 0x16)
@@ -404,6 +809,23 @@ namespace Hmu_HW5103 {
   /** included parts */
   union _includes {
     Hcmode_inc,
-    Errors_inc,
   }
+
+  enum Values_Domestic_hot_water_mod__eco_normal_balance_ {
+    eco: 0,
+    normal: 1,
+    balance: 2,
+  }
+
+  enum Values_immersion_heater_applications {
+    off: 0,
+    heating: 1,
+    dhw: 2,
+    both: 3,
+  }
+
+  enum Values_pumpperc {
+    auto: 49,
+  }
+
 }
--- a/src/vaillant/08.hmu.HW5103.tsp
+++ b/src/vaillant/08.hmu.HW5103.tsp
@@ -1,6 +1,7 @@
 import "@ebusd/ebus-typespec";
 import "./_templates.tsp";
 import "./hcmode_inc.tsp";
+import "./errors_inc.tsp";
 using Ebus;
 using Ebus.Num;
 using Ebus.Dtm;
@@ -756,40 +757,6 @@ namespace Hmu_HW5103 {
     value: tempv;
   }
 
-  // Error handling for this HW5103 variant.
-  //
-  // Errorhistory (b503 0101) is intentionally not exposed here. On this HMU
-  // the request returns a single byte 0x00 when no history entry is available,
-  // while the generic Errors_inc.Errorhistory definition expects a full
-  // status/time/date/error record and therefore fails with "invalid position".
-  // See ebusd-configuration issue #638.
-  // Currenterror and Clearerrorhistory are kept unchanged.
-
-  /** default *r for current errors */
-  @base(MF, 0x3, 0)
-  model r_error_current {}
-
-  /** current errors */
-  @inherit(r_error_current)
-  @ext(1)
-  model Currenterror {
-    value: errors;
-  }
-
-  /** default *wi for clearing error history */
-  @write
-  @auth("install")
-  @base(MF, 0x3, 0x2)
-  model wi_error_clear {}
-
-  /** clear error history */
-  @inherit(wi_error_clear)
-  @ext(1)
-  model Clearerrorhistory {
-    @in
-    cleared: yesno;
-  }
-
   // B516
   /** default *r */
   @base(MF, 0x16)
@@ -809,6 +776,7 @@ namespace Hmu_HW5103 {
   /** included parts */
   union _includes {
     Hcmode_inc,
+    Errors_inc,
   }
 
   enum Values_Domestic_hot_water_mod__eco_normal_balance_ {
--- a/src/vaillant/_templates.tsp
+++ b/src/vaillant/_templates.tsp
@@ -66,6 +66,10 @@ scalar twopointanalog extends UCH;
 @minValue(0) @maxValue(100) @step(0.5)
 scalar percentv extends EXP;
 
+/** percentage */
+@unit("%")
+scalar percent2 extends UIN;
+
 /** date */
 scalar date extends HDA3;
 
@@ -179,6 +183,10 @@ scalar calibrations extends SCH;
 @unit("K")
 @minValue(-50) @maxValue(50) @step(0.5)
 scalar calibrationv extends EXP;
+
+@unit("K")
+@divisor(16)
+scalar hysteresis extends UIN;
 
 @unit("°min")
 scalar integral extends SIN;


ClausLDK:
@john30 - all changed verified and tested with local upload

buliwyf42:
Thank you for putting this together — having the HW5103 work in one tested place is really helpful, and the Status07 and #607 parts look good to me.

One thing I'd like to add before the Errorhistory part goes in, because I think there may be a subtlety worth checking first.

## Errorhistory may be fixable rather than removable

`Errorhistory` has a master field `index` (`index,m,UCH`), so the request needs to carry it: `NN=03`, `DD=01 01 <index>`. The raw query quoted in #638 is `3108b503020101` — `NN=02`, without the index byte. When the request is incomplete the HMU answers with a single `00`, which looks exactly like "register not implemented".

This is genuinely easy to miss, because ebusd itself produces the same picture: a message with an `m` field can't be polled without an argument, so any integration config that pulls `Errorhistory` into the poll list (`filter-seen` in `mqtt-hassio.cfg`, for example) generates the same 1-byte reply and the same `invalid position` in the log. I ran into that on my own system a while back and drew the same conclusion at first.

With an index supplied, my HMU (`MF=Vaillant;ID=HMU00;SW=0902;HW=5103`) does return full 10-byte records. Raw log, unedited:

```
>3108b5030301010052<000a0200091411241600ff001b>00   index 0
>3108b5030301010153<000a0243121609241600ff0071>00   index 1
>3108b5030301010250<000a0247071209241600ff00f1>00   index 2
>3108b5030301010351<000a0206122604241600000047>00   index 3
>3108b5030301010456<000a0233142504241600000030>00   index 4
>3108b503030101095b<000a010000000000ffff000092>00   index 9
```

Payload split as `status(1) time(2) date(3) error(2) + 2 trailing`:

| index | status | time | date | error |
|---|---|---|---|---|
| 0 | `02` | `0009` | `141124` | `1600` |
| 1 | `02` | `4312` | `160924` | `1600` |
| 2 | `02` | `4707` | `120924` | `1600` |
| 3 | `02` | `0612` | `260424` | `1600` |
| 4 | `02` | `3314` | `250424` | `1600` |
| 9 | `01` | `0000` | `000000` | `ffff` |

`status=2` → entry present, `status=1` → empty slot, `error=0xffff` → none.

## What actually breaks the decode: the field types

The device sends **BCD**, while `errorhistory` in `_templates.tsp` uses the hex variants (`scalar time2 extends VTM`, `scalar date extends HDA3`). Checked with `ebusctl decode` against the bytes above:

| field | current type | bytes | result | BCD type | result |
|---|---|---|---|---|---|
| time | `VTM` | `4312` | `ERR: argument value out of valid range` | `BTM` | `12:43` |
| date | `HDA:3` | `141124` | `ERR: argument value out of valid range` | `BDA:3` | `14.11.2024` |
| date | `HDA:3` | `160924` | `22.09.2036` — no error, just wrong | `BDA:3` | `16.09.2024` |

The last row is the one that worries me a little: `HDA:3` doesn't fail there, it returns a plausible-looking but incorrect date. `error: UIN` is fine (`1600` → 22), and the two trailing bytes are ignored by ebusd. This holds across all seven indices I read.

So a HW5103-local override of those two field types might get the message working instead of dropping it. I'm happy to prepare that as a separate small PR against `08.hmu.HW5103.tsp` if that's useful, so it doesn't hold up the rest of #660 — whatever fits your plans best.

**One caveat I should be upfront about:** #638 is on `SW=0522` and I'm on `SW=0902`, and @chrizzzp already found in that same issue that `PowerConsumptionHmu` only exists from `SW>=0901`. So I can't rule out that 0522 really has no error history, and my measurements may simply not transfer. @MarkBovee, if you have a moment, would you mind re-testing with the index byte included — `3108b50303010100`? That would settle it either way, and if 0522 still answers `00`, removing the message for HW5103 makes complete sense to me.

## One small question

`BuildingCircuitFlow`: the `l/h` → `L/h` change is a unit rename on an existing register, so it will show up as a changed unit on already-configured Home Assistant sensors. Intentional (SI capitalisation), or incidental? Either is fine, just wanted to mention it.

## Status07 / #614

Status07 here is #614 as-is — that's completely fine by me, and thank you for the credit. Just so it doesn't get merged twice, @john30: #614 and this PR now carry the same model, so whichever lands first will make the other conflict. I'm glad to close #614 if #660 is the better path forward.

Clauslarsen75:
> Thank you for putting this together — having the HW5103 work in one tested place is really helpful, and the Status07 and #607 parts look good to me.
> 
> One thing I'd like to add before the Errorhistory part goes in, because I think there may be a subtlety worth checking first.
> 
> ## Errorhistory may be fixable rather than removable
> `Errorhistory` has a master field `index` (`index,m,UCH`), so the request needs to carry it: `NN=03`, `DD=01 01 <index>`. The raw query quoted in #638 is `3108b503020101` — `NN=02`, without the index byte. When the request is incomplete the HMU answers with a single `00`, which looks exactly like "register not implemented".
> 
> This is genuinely easy to miss, because ebusd itself produces the same picture: a message with an `m` field can't be polled without an argument, so any integration config that pulls `Errorhistory` into the poll list (`filter-seen` in `mqtt-hassio.cfg`, for example) generates the same 1-byte reply and the same `invalid position` in the log. I ran into that on my own system a while back and drew the same conclusion at first.
> 
> With an index supplied, my HMU (`MF=Vaillant;ID=HMU00;SW=0902;HW=5103`) does return full 10-byte records. Raw log, unedited:
> 
> ```
> >3108b5030301010052<000a0200091411241600ff001b>00   index 0
> >3108b5030301010153<000a0243121609241600ff0071>00   index 1
> >3108b5030301010250<000a0247071209241600ff00f1>00   index 2
> >3108b5030301010351<000a0206122604241600000047>00   index 3
> >3108b5030301010456<000a0233142504241600000030>00   index 4
> >3108b503030101095b<000a010000000000ffff000092>00   index 9
> ```
> 
> Payload split as `status(1) time(2) date(3) error(2) + 2 trailing`:
> 
> index	status	time	date	error
> 0	`02`	`0009`	`141124`	`1600`
> 1	`02`	`4312`	`160924`	`1600`
> 2	`02`	`4707`	`120924`	`1600`
> 3	`02`	`0612`	`260424`	`1600`
> 4	`02`	`3314`	`250424`	`1600`
> 9	`01`	`0000`	`000000`	`ffff`
> `status=2` → entry present, `status=1` → empty slot, `error=0xffff` → none.
> 
> ## What actually breaks the decode: the field types
> The device sends **BCD**, while `errorhistory` in `_templates.tsp` uses the hex variants (`scalar time2 extends VTM`, `scalar date extends HDA3`). Checked with `ebusctl decode` against the bytes above:
> 
> field	current type	bytes	result	BCD type	result
> time	`VTM`	`4312`	`ERR: argument value out of valid range`	`BTM`	`12:43`
> date	`HDA:3`	`141124`	`ERR: argument value out of valid range`	`BDA:3`	`14.11.2024`
> date	`HDA:3`	`160924`	`22.09.2036` — no error, just wrong	`BDA:3`	`16.09.2024`
> The last row is the one that worries me a little: `HDA:3` doesn't fail there, it returns a plausible-looking but incorrect date. `error: UIN` is fine (`1600` → 22), and the two trailing bytes are ignored by ebusd. This holds across all seven indices I read.
> 
> So a HW5103-local override of those two field types might get the message working instead of dropping it. I'm happy to prepare that as a separate small PR against `08.hmu.HW5103.tsp` if that's useful, so it doesn't hold up the rest of #660 — whatever fits your plans best.
> 
> **One caveat I should be upfront about:** #638 is on `SW=0522` and I'm on `SW=0902`, and @chrizzzp already found in that same issue that `PowerConsumptionHmu` only exists from `SW>=0901`. So I can't rule out that 0522 really has no error history, and my measurements may simply not transfer. @MarkBovee, if you have a moment, would you mind re-testing with the index byte included — `3108b50303010100`? That would settle it either way, and if 0522 still answers `00`, removing the message for HW5103 makes complete sense to me.
> 
> ## One small question
> `BuildingCircuitFlow`: the `l/h` → `L/h` change is a unit rename on an existing register, so it will show up as a changed unit on already-configured Home Assistant sensors. Intentional (SI capitalisation), or incidental? Either is fine, just wanted to mention it.
> 
> ## Status07 / #614
> Status07 here is #614 as-is — that's completely fine by me, and thank you for the credit. Just so it doesn't get merged twice, @john30: #614 and this PR now carry the same model, so whichever lands first will make the other conflict. I'm glad to close #614 if #660 is the better path forward.

Thanks for the detailed analysis — I tested this on my own HW5103 before replying.

My system is:

`MF=Vaillant; ID=HMU00; SW=0905; HW=5103`

I verified the identification directly from the bus:

`b5484d55303009055103`

I then tested Errorhistory with the index byte included, using the same source address as in your examples:

`hex -s 31 08b50303010100`
`hex -s 31 08b50303010101`
`hex -s 31 08b50303010109`

All three requests complete successfully, but return no data payload on my SW0905.

So this is interestingly different from your HW5103/SW0902, which returns the full 10-byte records with the indexed request.

That suggests Errorhistory behaviour may be software-version dependent even within HW5103, rather than purely hardware dependent.

I agree that removing Errorhistory for all HW5103 devices would therefore be too broad. It may make more sense to keep the definition and handle the supported SW range separately, if we can establish where the behaviour changes.

Regarding `BuildingCircuitFlow`: the `l/h` → `L/h` change was intentional. The actual value and scaling are unchanged; this was only to use the Home Assistant-compatible unit spelling.

Thanks again for catching the indexed Errorhistory detail — it revealed a useful SW0902 vs SW0905 difference.

ClausLDK:
Thanks for the detailed analysis — I tested this on my own HW5103 before
replying.

My system is:

MF=Vaillant; ID=HMU00; SW=0905; HW=5103

I verified the identification directly from the bus:

b5484d55303009055103

I then tested Errorhistory with the index byte included, using the same
source address as in your examples:

hex -s 31 08b50303010100
hex -s 31 08b50303010101
hex -s 31 08b50303010109

All three requests complete successfully, but return no data payload on my
SW0905.

So this is interestingly different from your HW5103/SW0902, which returns
the full 10-byte records with the indexed request.

That suggests Errorhistory behaviour may be software-version dependent even
within HW5103, rather than purely hardware dependent.

Based on this, I have removed the HW5103-specific Errorhistory workaround
from this PR and restored the generic Errors_inc include. I agree that
removing Errorhistory for all HW5103 devices would be too broad; it
probably makes more sense to investigate the supported SW range separately.

Regarding BuildingCircuitFlow: the l/h → L/h change was intentional. The
actual value and scaling are unchanged; this was only to use the Home
Assistant-compatible unit spelling.

Thanks again for catching the indexed Errorhistory detail — it revealed a
useful SW0902 vs SW0905 difference.

On Sat, 05 Sep 2026 04:07:15 -0700, Dominik Sennfelder
***@***.*** wrote:

buliwyf42 left a comment (john30/ebusd-configuration#660)
<https://github.com/john30/ebusd-configuration/pull/660#issuecomment-5551349406>

Thank you for putting this together — having the HW5103 work in one tested
place is really helpful, and the Status07 and #607
<https://github.com/john30/ebusd-configuration/pull/607> parts look good to
me.

One thing I’d like to add before the Errorhistory part goes in, because I
think there may be a subtlety worth checking first.

Errorhistory may be fixable rather than removable

Errorhistory has a master field index (index,m,UCH), so the request needs
to carry it: NN=03, DD=01 01 . The raw query quoted in #638
<https://github.com/john30/ebusd-configuration/issues/638> is
3108b503020101 — NN=02, without the index byte. When the request is
incomplete the HMU answers with a single 00, which looks exactly like
“register not implemented”.

This is genuinely easy to miss, because ebusd itself produces the same
picture: a message with an m field can’t be polled without an argument, so
any integration config that pulls Errorhistory into the poll list
(filter-seen in mqtt-hassio.cfg, for example) generates the same 1-byte
reply and the same invalid position in the log. I ran into that on my own
system a while back and drew the same conclusion at first.

With an index supplied, my HMU (MF=Vaillant;ID=HMU00;SW=0902;HW=5103) does
return full 10-byte records. Raw log, unedited:

3108b5030301010052<000a0200091411241600ff001b>00 index 0
3108b5030301010153<000a0243121609241600ff0071>00 index 1
3108b5030301010250<000a0247071209241600ff00f1>00 index 2
3108b5030301010351<000a0206122604241600000047>00 index 3
3108b5030301010456<000a0233142504241600000030>00 index 4
3108b503030101095b<000a010000000000ffff000092>00 index 9

Payload split as status(1) time(2) date(3) error(2) + 2 trailing:

index
status
time
date
error

0
02
0009
141124
1600

1
02
4312
160924
1600

2
02
4707
120924
1600

3
02
0612
260424
1600

4
02
3314
250424
1600

9
01
0000
000000
ffff

status=2 → entry present, status=1 → empty slot, error=0xffff → none.

What actually breaks the decode: the field types

The device sends BCD, while errorhistory in _templates.tsp uses the hex
variants (scalar time2 extends VTM, scalar date extends HDA3). Checked with
ebusctl decode against the bytes above:

field
current type
bytes
result
BCD type
result

time
VTM
4312
ERR: argument value out of valid range
BTM
12:43

date
HDA:3
141124
ERR: argument value out of valid range
BDA:3
14.11.2024

date
HDA:3
160924
22.09.2036 — no error, just wrong
BDA:3
16.09.2024

The last row is the one that worries me a little: HDA:3 doesn’t fail there,
it returns a plausible-looking but incorrect date. error: UIN is fine (1600
→ 22), and the two trailing bytes are ignored by ebusd. This holds across
all seven indices I read.

So a HW5103-local override of those two field types might get the message
working instead of dropping it. I’m happy to prepare that as a separate
small PR against 08.hmu.HW5103.tsp if that’s useful, so it doesn’t hold up
the rest of #660 <https://github.com/john30/ebusd-configuration/pull/660> —
whatever fits your plans best.

One caveat I should be upfront about: #638
<https://github.com/john30/ebusd-configuration/issues/638> is on SW=0522
and I’m on SW=0902, and @chrizzzp <https://github.com/chrizzzp> already
found in that same issue that PowerConsumptionHmu only exists from
SW>=0901. So I can’t rule out that 0522 really has no error history, and my
measurements may simply not transfer. @MarkBovee
<https://github.com/MarkBovee>, if you have a moment, would you mind
re-testing with the index byte included — 3108b50303010100? That would
settle it either way, and if 0522 still answers 00, removing the message
for HW5103 makes complete sense to me.

One small question

BuildingCircuitFlow: the l/h → L/h change is a unit rename on an existing
register, so it will show up as a changed unit on already-configured Home
Assistant sensors. Intentional (SI capitalisation), or incidental? Either
is fine, just wanted to mention it.

Status07 / #614 <https://github.com/john30/ebusd-configuration/pull/614>

Status07 here is #614
<https://github.com/john30/ebusd-configuration/pull/614> as-is — that’s
completely fine by me, and thank you for the credit. Just so it doesn’t get
merged twice, @john30 <https://github.com/john30>: #614
<https://github.com/john30/ebusd-configuration/pull/614> and this PR now
carry the same model, so whichever lands first will make the other
conflict. I’m glad to close #614
<https://github.com/john30/ebusd-configuration/pull/614> if #660
<https://github.com/john30/ebusd-configuration/pull/660> is the better path
forward.

ClausLDK:
> Thank you for putting this together — having the HW5103 work in one tested place is really helpful, and the Status07 and #607 parts look good to me.
> 
> One thing I'd like to add before the Errorhistory part goes in, because I think there may be a subtlety worth checking first.
> 
> ## Errorhistory may be fixable rather than removable
> `Errorhistory` has a master field `index` (`index,m,UCH`), so the request needs to carry it: `NN=03`, `DD=01 01 <index>`. The raw query quoted in #638 is `3108b503020101` — `NN=02`, without the index byte. When the request is incomplete the HMU answers with a single `00`, which looks exactly like "register not implemented".
> 
> This is genuinely easy to miss, because ebusd itself produces the same picture: a message with an `m` field can't be polled without an argument, so any integration config that pulls `Errorhistory` into the poll list (`filter-seen` in `mqtt-hassio.cfg`, for example) generates the same 1-byte reply and the same `invalid position` in the log. I ran into that on my own system a while back and drew the same conclusion at first.
> 
> With an index supplied, my HMU (`MF=Vaillant;ID=HMU00;SW=0902;HW=5103`) does return full 10-byte records. Raw log, unedited:
> 
> ```
> >3108b5030301010052<000a0200091411241600ff001b>00   index 0
> >3108b5030301010153<000a0243121609241600ff0071>00   index 1
> >3108b5030301010250<000a0247071209241600ff00f1>00   index 2
> >3108b5030301010351<000a0206122604241600000047>00   index 3
> >3108b5030301010456<000a0233142504241600000030>00   index 4
> >3108b503030101095b<000a010000000000ffff000092>00   index 9
> ```
> 
> Payload split as `status(1) time(2) date(3) error(2) + 2 trailing`:
> 
> index	status	time	date	error
> 0	`02`	`0009`	`141124`	`1600`
> 1	`02`	`4312`	`160924`	`1600`
> 2	`02`	`4707`	`120924`	`1600`
> 3	`02`	`0612`	`260424`	`1600`
> 4	`02`	`3314`	`250424`	`1600`
> 9	`01`	`0000`	`000000`	`ffff`
> `status=2` → entry present, `status=1` → empty slot, `error=0xffff` → none.
> 
> ## What actually breaks the decode: the field types
> The device sends **BCD**, while `errorhistory` in `_templates.tsp` uses the hex variants (`scalar time2 extends VTM`, `scalar date extends HDA3`). Checked with `ebusctl decode` against the bytes above:
> 
> field	current type	bytes	result	BCD type	result
> time	`VTM`	`4312`	`ERR: argument value out of valid range`	`BTM`	`12:43`
> date	`HDA:3`	`141124`	`ERR: argument value out of valid range`	`BDA:3`	`14.11.2024`
> date	`HDA:3`	`160924`	`22.09.2036` — no error, just wrong	`BDA:3`	`16.09.2024`
> The last row is the one that worries me a little: `HDA:3` doesn't fail there, it returns a plausible-looking but incorrect date. `error: UIN` is fine (`1600` → 22), and the two trailing bytes are ignored by ebusd. This holds across all seven indices I read.
> 
> So a HW5103-local override of those two field types might get the message working instead of dropping it. I'm happy to prepare that as a separate small PR against `08.hmu.HW5103.tsp` if that's useful, so it doesn't hold up the rest of #660 — whatever fits your plans best.
> 
> **One caveat I should be upfront about:** #638 is on `SW=0522` and I'm on `SW=0902`, and @chrizzzp already found in that same issue that `PowerConsumptionHmu` only exists from `SW>=0901`. So I can't rule out that 0522 really has no error history, and my measurements may simply not transfer. @MarkBovee, if you have a moment, would you mind re-testing with the index byte included — `3108b50303010100`? That would settle it either way, and if 0522 still answers `00`, removing the message for HW5103 makes complete sense to me.
> 
> ## One small question
> `BuildingCircuitFlow`: the `l/h` → `L/h` change is a unit rename on an existing register, so it will show up as a changed unit on already-configured Home Assistant sensors. Intentional (SI capitalisation), or incidental? Either is fine, just wanted to mention it.
> 
> ## Status07 / #614
> Status07 here is #614 as-is — that's completely fine by me, and thank you for the credit. Just so it doesn't get merged twice, @john30: #614 and this PR now carry the same model, so whichever lands first will make the other conflict. I'm glad to close #614 if #660 is the better path forward.

Thanks for the detailed analysis — I tested this on my own HW5103 before replying.

My system is:

`MF=Vaillant; ID=HMU00; SW=0905; HW=5103`

I verified the identification directly from the bus:

`b5484d55303009055103`

I then tested Errorhistory with the index byte included, using the same source address as in your examples:

`hex -s 31 08b50303010100`
`hex -s 31 08b50303010101`
`hex -s 31 08b50303010109`

All three requests complete successfully, but return no data payload on my SW0905.

So this is interestingly different from your HW5103/SW0902, which returns the full 10-byte records with the indexed request.

That suggests Errorhistory behaviour may be software-version dependent even within HW5103, rather than purely hardware dependent.

I agree that removing Errorhistory for all HW5103 devices would therefore be too broad. It may make more sense to keep the definition and handle the supported SW range separately, if we can establish where the behaviour changes.

Regarding `BuildingCircuitFlow`: the `l/h` → `L/h` change was intentional. The actual value and scaling are unchanged; this was only to use the Home Assistant-compatible unit spelling.

Thanks again for catching the indexed Errorhistory detail — it revealed a useful SW0902 vs SW0905 difference.

ClausLDK:
Verified HwcMode read/write on a real HMU00, HW5103, SW0905 system.

Register:

B51A 05/06 FF 34 44

Read values were confirmed as:

   - Eco → 0
   - Normal → 1
   - Balance → 2

Write access was also tested successfully in both directions:

   - Normal → Balance
   - Balance → Normal

Each write was verified by reading the value back from the HMU.

So the @ext(0x34, 0x44) definition and the eco/normal/balance enum in this
PR are confirmed for both read and write access on SW0905.

This addresses #562.

On Sat, 05 Sep 2026 04:07:15 -0700, Dominik Sennfelder
***@***.*** wrote:

buliwyf42 left a comment (john30/ebusd-configuration#660)
<https://github.com/john30/ebusd-configuration/pull/660#issuecomment-5551349406>

Thank you for putting this together — having the HW5103 work in one tested
place is really helpful, and the Status07 and #607
<https://github.com/john30/ebusd-configuration/pull/607> parts look good to
me.

One thing I’d like to add before the Errorhistory part goes in, because I
think there may be a subtlety worth checking first.

Errorhistory may be fixable rather than removable

Errorhistory has a master field index (index,m,UCH), so the request needs
to carry it: NN=03, DD=01 01 . The raw query quoted in #638
<https://github.com/john30/ebusd-configuration/issues/638> is
3108b503020101 — NN=02, without the index byte. When the request is
incomplete the HMU answers with a single 00, which looks exactly like
“register not implemented”.

This is genuinely easy to miss, because ebusd itself produces the same
picture: a message with an m field can’t be polled without an argument, so
any integration config that pulls Errorhistory into the poll list
(filter-seen in mqtt-hassio.cfg, for example) generates the same 1-byte
reply and the same invalid position in the log. I ran into that on my own
system a while back and drew the same conclusion at first.

With an index supplied, my HMU (MF=Vaillant;ID=HMU00;SW=0902;HW=5103) does
return full 10-byte records. Raw log, unedited:

3108b5030301010052<000a0200091411241600ff001b>00 index 0
3108b5030301010153<000a0243121609241600ff0071>00 index 1
3108b5030301010250<000a0247071209241600ff00f1>00 index 2
3108b5030301010351<000a0206122604241600000047>00 index 3
3108b5030301010456<000a0233142504241600000030>00 index 4
3108b503030101095b<000a010000000000ffff000092>00 index 9

Payload split as status(1) time(2) date(3) error(2) + 2 trailing:

index
status
time
date
error

0
02
0009
141124
1600

1
02
4312
160924
1600

2
02
4707
120924
1600

3
02
0612
260424
1600

4
02
3314
250424
1600

9
01
0000
000000
ffff

status=2 → entry present, status=1 → empty slot, error=0xffff → none.

What actually breaks the decode: the field types

The device sends BCD, while errorhistory in _templates.tsp uses the hex
variants (scalar time2 extends VTM, scalar date extends HDA3). Checked with
ebusctl decode against the bytes above:

field
current type
bytes
result
BCD type
result

time
VTM
4312
ERR: argument value out of valid range
BTM
12:43

date
HDA:3
141124
ERR: argument value out of valid range
BDA:3
14.11.2024

date
HDA:3
160924
22.09.2036 — no error, just wrong
BDA:3
16.09.2024

The last row is the one that worries me a little: HDA:3 doesn’t fail there,
it returns a plausible-looking but incorrect date. error: UIN is fine (1600
→ 22), and the two trailing bytes are ignored by ebusd. This holds across
all seven indices I read.

So a HW5103-local override of those two field types might get the message
working instead of dropping it. I’m happy to prepare that as a separate
small PR against 08.hmu.HW5103.tsp if that’s useful, so it doesn’t hold up
the rest of #660 <https://github.com/john30/ebusd-configuration/pull/660> —
whatever fits your plans best.

One caveat I should be upfront about: #638
<https://github.com/john30/ebusd-configuration/issues/638> is on SW=0522
and I’m on SW=0902, and @chrizzzp <https://github.com/chrizzzp> already
found in that same issue that PowerConsumptionHmu only exists from
SW>=0901. So I can’t rule out that 0522 really has no error history, and my
measurements may simply not transfer. @MarkBovee
<https://github.com/MarkBovee>, if you have a moment, would you mind
re-testing with the index byte included — 3108b50303010100? That would
settle it either way, and if 0522 still answers 00, removing the message
for HW5103 makes complete sense to me.

One small question

BuildingCircuitFlow: the l/h → L/h change is a unit rename on an existing
register, so it will show up as a changed unit on already-configured Home
Assistant sensors. Intentional (SI capitalisation), or incidental? Either
is fine, just wanted to mention it.

Status07 / #614 <https://github.com/john30/ebusd-configuration/pull/614>

Status07 here is #614
<https://github.com/john30/ebusd-configuration/pull/614> as-is — that’s
completely fine by me, and thank you for the credit. Just so it doesn’t get
merged twice, @john30 <https://github.com/john30>: #614
<https://github.com/john30/ebusd-configuration/pull/614> and this PR now
carry the same model, so whichever lands first will make the other
conflict. I’m glad to close #614
<https://github.com/john30/ebusd-configuration/pull/614> if #660
<https://github.com/john30/ebusd-configuration/pull/660> is the better path
forward.

ClausLDK:
Also verified RunDataElectricPowerConsumption on a real HMU00, HW5103,
SW0905 system.

Register:

B509 54 02 00 5B 0D

The HW5103 responds with the expected 4-byte header followed by an EXP
value in watts.

Observed examples:

   - 02015b0d0000d040 → about 6.5 W
   - 02015b0d84a48342 → about 65.8 W

The value changes with system activity and the payload format matches the
generic Vaillant definition using IGN:4 + EXP with unit W.

I therefore added RunDataElectricPowerConsumption to the HW5103 B509
section in this PR and verified that the TypeSpec compiles successfully and
generates:

b509,5402005b0d,ign,,IGN:4,...,value,,EXP,,,W

This addresses #595.

On Sat, 05 Sep 2026 04:07:15 -0700, Dominik Sennfelder
***@***.*** wrote:

buliwyf42 left a comment (john30/ebusd-configuration#660)
<https://github.com/john30/ebusd-configuration/pull/660#issuecomment-5551349406>

Thank you for putting this together — having the HW5103 work in one tested
place is really helpful, and the Status07 and #607
<https://github.com/john30/ebusd-configuration/pull/607> parts look good to
me.

One thing I’d like to add before the Errorhistory part goes in, because I
think there may be a subtlety worth checking first.

Errorhistory may be fixable rather than removable

Errorhistory has a master field index (index,m,UCH), so the request needs
to carry it: NN=03, DD=01 01 . The raw query quoted in #638
<https://github.com/john30/ebusd-configuration/issues/638> is
3108b503020101 — NN=02, without the index byte. When the request is
incomplete the HMU answers with a single 00, which looks exactly like
“register not implemented”.

This is genuinely easy to miss, because ebusd itself produces the same
picture: a message with an m field can’t be polled without an argument, so
any integration config that pulls Errorhistory into the poll list
(filter-seen in mqtt-hassio.cfg, for example) generates the same 1-byte
reply and the same invalid position in the log. I ran into that on my own
system a while back and drew the same conclusion at first.

With an index supplied, my HMU (MF=Vaillant;ID=HMU00;SW=0902;HW=5103) does
return full 10-byte records. Raw log, unedited:

3108b5030301010052<000a0200091411241600ff001b>00 index 0
3108b5030301010153<000a0243121609241600ff0071>00 index 1
3108b5030301010250<000a0247071209241600ff00f1>00 index 2
3108b5030301010351<000a0206122604241600000047>00 index 3
3108b5030301010456<000a0233142504241600000030>00 index 4
3108b503030101095b<000a010000000000ffff000092>00 index 9

Payload split as status(1) time(2) date(3) error(2) + 2 trailing:

index
status
time
date
error

0
02
0009
141124
1600

1
02
4312
160924
1600

2
02
4707
120924
1600

3
02
0612
260424
1600

4
02
3314
250424
1600

9
01
0000
000000
ffff

status=2 → entry present, status=1 → empty slot, error=0xffff → none.

What actually breaks the decode: the field types

The device sends BCD, while errorhistory in _templates.tsp uses the hex
variants (scalar time2 extends VTM, scalar date extends HDA3). Checked with
ebusctl decode against the bytes above:

field
current type
bytes
result
BCD type
result

time
VTM
4312
ERR: argument value out of valid range
BTM
12:43

date
HDA:3
141124
ERR: argument value out of valid range
BDA:3
14.11.2024

date
HDA:3
160924
22.09.2036 — no error, just wrong
BDA:3
16.09.2024

The last row is the one that worries me a little: HDA:3 doesn’t fail there,
it returns a plausible-looking but incorrect date. error: UIN is fine (1600
→ 22), and the two trailing bytes are ignored by ebusd. This holds across
all seven indices I read.

So a HW5103-local override of those two field types might get the message
working instead of dropping it. I’m happy to prepare that as a separate
small PR against 08.hmu.HW5103.tsp if that’s useful, so it doesn’t hold up
the rest of #660 <https://github.com/john30/ebusd-configuration/pull/660> —
whatever fits your plans best.

One caveat I should be upfront about: #638
<https://github.com/john30/ebusd-configuration/issues/638> is on SW=0522
and I’m on SW=0902, and @chrizzzp <https://github.com/chrizzzp> already
found in that same issue that PowerConsumptionHmu only exists from
SW>=0901. So I can’t rule out that 0522 really has no error history, and my
measurements may simply not transfer. @MarkBovee
<https://github.com/MarkBovee>, if you have a moment, would you mind
re-testing with the index byte included — 3108b50303010100? That would
settle it either way, and if 0522 still answers 00, removing the message
for HW5103 makes complete sense to me.

One small question

BuildingCircuitFlow: the l/h → L/h change is a unit rename on an existing
register, so it will show up as a changed unit on already-configured Home
Assistant sensors. Intentional (SI capitalisation), or incidental? Either
is fine, just wanted to mention it.

Status07 / #614 <https://github.com/john30/ebusd-configuration/pull/614>

Status07 here is #614
<https://github.com/john30/ebusd-configuration/pull/614> as-is — that’s
completely fine by me, and thank you for the credit. Just so it doesn’t get
merged twice, @john30 <https://github.com/john30>: #614
<https://github.com/john30/ebusd-configuration/pull/614> and this PR now
carry the same model, so whichever lands first will make the other
conflict. I’m glad to close #614
<https://github.com/john30/ebusd-configuration/pull/614> if #660
<https://github.com/john30/ebusd-configuration/pull/660> is the better path
forward.

ClausLDK:
### Additional HW5103 telemetry

The PR also adds two additional read-only values, both live-verified on:

`MF=Vaillant; ID=HMU00; SW=0905; HW=5103`

#### Air intake temperature

Adds:

`AirIntakeTemp`

Register:

`B51A 05 FF 32 26`

Live test:

`31 08 B51A 05 FF 32 26`

Response:

`ff083e28010000000000`

Decoded value:

`18.5 °C`

Generated CSV uses `D2C` with unit `°C`.

#### Cooling operating hours

Adds:

`RunStatsCoolHours`

Register:

`B509 540200 BB 0B`

Live test:

`31 08 B509 540200 BB 0B`

Response:

`0201bb0b5b020000`

Decoded value:

`603 h`

This complements the existing B509 runtime statistics:

* `RunStatsHMUHours` (`B8 0B`)
* `RunStatsHcHours` (`B9 0B`)
* `RunStatsHwcHours` (`BC 0B`)

During verification of PR #604, the corresponding B51A `Hours*` values were also tested:

* `Hours` = 7738 h
* `HoursHc` = 4419 h
* `HoursCool` = 603 h
* `HoursHwc` = 879 h

`Hours`, `HoursHc` and `HoursHwc` were found to be exact duplicates of the existing B509 `RunStats*` values and were therefore not added here.

For cooling hours, the corresponding B509 register was identified as `BB 0B`, returning exactly the same `603 h` as the B51A `HoursCool` value. Therefore `RunStatsCoolHours` was added to the existing B509 RunStats family instead of adding the duplicate B51A `HoursCool` message.

buliwyf42:
Tested, as promised. Thanks for your patience.

**System:** `MF=Vaillant; ID=HMU00; SW=0902; HW=5103` — a cascade of two aroTHERM plus VWL 75/6 A S2 behind a VR32b coupler, sharing one VWZ MEH 97/6 hydraulic station. ebusd 26.1.26.1.

I compiled this branch, merged the 56 new CSV lines line-wise into my local `08.hmu.HW5103.csv` (leaving out `Status07`, which I already carry, and the `l/h → L/h` change), and read all 30 `r` registers. Everything was rolled back afterwards.

## 16 registers return plausible values

| Register | Value | Slave payload |
|---|---|---|
| `AirIntakeTemp` | 18.62 °C | `ff083e 2a01 0000 0000 00` |
| `RunDataElectricPowerConsumption` | 8.5 W | |
| `RunStatsCoolHours` | 35 h | |
| `HwcMode` | normal | `ff0200 0100 0000 0200 01` |
| `CompHysteresisHeating` | 7.00 K | `ff0207 7000 3000 f000 10` |
| `CompHysteresisCooling` | 5.00 K | `ff0207 5000 3000 f000 10` |
| `CompStartHeatingFrom` | −60 °min | `ff0225 c4ff 9cff e2ff 05` |
| `CompStartCoolingFrom` | 60 °min | `ff0225 3c00 1e00 6400 05` |
| `MaxRemainingDeltaP` | 900 | `ff022d 8403 c800 8403 0a` |
| `MaximumS21BlockAllowed` | 5 h | `ff020d 0500 0000 0900 01` |
| `NoiseReductionLevel` | 40 % | `ff021b 2800 1e00 3c00 01` |
| `BlockTimeAfterRestart` | 0 min | `ff0234 0000 0000 7800 01` |
| `BuildingCircuitPumpOutputHeating` | auto | `ff0240 3100 3100 6400 01` |
| `BuildingCircuitPumpOutputCooling` | auto | `ff0240 3100 3100 6400 01` |
| `BuildingCircuitPumpOutputHwc` | 100 % | `ff0240 6400 3100 6400 01` |
| `HMUSystemStarts` | 23 | `ff0802 1700 0000 0000 00` |

`RunStatsCoolHours` = 35 h is a good independent check: this system only started cooling at the end of August 2026 (the function was retrofitted), so a small number is exactly right. Your 603 h and my 35 h are consistent with two very different histories.

## 13 registers answer with a 2-byte refusal

All identical: `NN=02`, payload **`ff 01`**.

`HeatCurve` (`352d`), `SummerSwitchOffTemp` (`352e`), `HcBivalencePoint` (`352f`), `HwcBivalencePoint` (`3530`), `HcAlternativePoint` (`3531`), `MaxFlowTemp` (`3532`), `MinFlowTemp` (`3533`), `HcEnabled` (`3534`), `HwcEnabled` (`3535`), `EmergencyModeImmersionHeaterEnabled` (`3537`), `HwcChargeHysteresis` (`3539`), `CoolTargetFlowTemp` (`352b`), `ImmersionHeaterEnabled` (`331e`)

```
>3108b51a0405ff352db6<0002ff0156>00
```

## The `b51a 05ff xxxx` response layout

Comparing the two groups makes the format readable, which I think is worth recording:

```
ff <flag> <sub> | <value LE 2B> | <min LE 2B> | <max LE 2B> | <step 1B>
```

* `flag = 02` → writable parameter, value block follows
* `flag = 08` → read-only value, min/max/step all zero (`AirIntakeTemp`, `HMUSystemStarts`)
* `flag = 01` → **parameter not present on this device**, reply ends after 2 bytes

That explains the `IGN:3` in the definitions, and it means the device hands you min/max/step for free. Cross-check: `CompHysteresisHeating` reports min 3.00, max 15.00, step 1.00 — exactly the `@minValue(3)` / `@maxValue(15)` in your definition.

I'd read the `ff 01` group as "not implemented on this installation" rather than a definition problem — so no change requested. My best guess is that on a cascade the heating-curve and bivalence parameters live in the controller rather than in the HMU, but that is a guess, not a measurement.

## Where the device disagrees with the annotations

| Register | Device reports | Definition |
|---|---|---|
| `CompStartHeatingFrom` | min −100, max −30 | `@minValue(-999) @maxValue(9)` |
| `CompStartCoolingFrom` | min 30, max 100 | `@minValue(0) @maxValue(999)` |
| `NoiseReductionLevel` | min 30, max 60 | `@minValue(40) @maxValue(60)` |

Only affects write validation. Interesting that `NoiseReductionLevel` differs between our units: your `@Ebus.example` payload carries min `2800` = 40, mine reports `1e00` = 30 — so that bound looks installation-dependent and may be better left off.

## One collision worth knowing about

`ActiveCoolingEnabled` = `b51a 05ff331b` is the same register I have been carrying since 2026-08-29 under the name `CoolingTechnology`. I found it when the VWZ (master `71`) wrote `b51a 06 <seq> 331b 0100` to the HMU at the exact moment the setting was changed at the indoor unit display. I read it as `UIN` with `0=none;1=active` plus the min/max fields; you read the first byte as `UCH` with `0=off;1=on`. Same register, same meaning — just two names.

That only bit me because both lines were in one file during the test:

```
[main error] error reading scan config file vaillant/08.hmu.HW5103.csv for ID "hmu00",
  SW0902, HW5103: ERR: duplicate entry, …:145: ERR: duplicate entry, duplicate ID
[main error] scan config 08: ERR: duplicate entry
```

Not a problem for your PR as such, since `CoolingTechnology` is not upstream — but if it ever goes upstream, one of the two names has to give way. Worth noting that ebusd's `--checkconfig` does **not** catch this; it counted all messages happily and the collision only surfaced when the scan config was loaded at runtime, aborting the rest of the file.

ClausLDK:
Thanks for the very thorough test — especially the min/max/step observation.

I checked the same three registers on my own system:

MF=Vaillant; ID=HMU00; SW=0905; HW=5103

Results:

   - CompStartHeatingFrom (341e): value -100, min -100, max -30, step 5
   - CompStartCoolingFrom (3421): value 100, min 30, max 100, step 5
   - NoiseReductionLevel (3428): value 60%, min 40%, max 60%, step 1

So the two integral bounds are identical on SW0902 and SW0905, which
confirms that the current annotations are too broad. I will change them to:

   - CompStartHeatingFrom: @minValue(-100), @maxValue(-30), @step(5)
   - CompStartCoolingFrom: @minValue(30), @maxValue(100), @step(5)

For NoiseReductionLevel, your SW0902 reports a minimum of 30% while my
SW0905 reports 40%. Since that bound is clearly installation/firmware
dependent, I will remove the fixed @minValue(40) and keep the common
@maxValue(60).

Your independent confirmation of AirIntakeTemp,
RunDataElectricPowerConsumption and RunStatsCoolHours is also very useful —
thanks again.

On Fri, 11 Sep 2026 03:22:15 -0700, Dominik Sennfelder
***@***.*** wrote:

buliwyf42 left a comment (john30/ebusd-configuration#660)
<https://github.com/john30/ebusd-configuration/pull/660#issuecomment-5633021475>

Tested, as promised. Thanks for your patience.

System: MF=Vaillant; ID=HMU00; SW=0902; HW=5103 — a cascade of two aroTHERM
plus VWL 75/6 A S2 behind a VR32b coupler, sharing one VWZ MEH 97/6
hydraulic station. ebusd 26.1.26.1.

I compiled this branch, merged the 56 new CSV lines line-wise into my local
08.hmu.HW5103.csv (leaving out Status07, which I already carry, and the l/h
→ L/h change), and read all 30 r registers. Everything was rolled back
afterwards.

16 registers return plausible values

Register
Value
Slave payload

AirIntakeTemp
18.62 °C
ff083e 2a01 0000 0000 00

RunDataElectricPowerConsumption
8.5 W

RunStatsCoolHours
35 h

HwcMode
normal
ff0200 0100 0000 0200 01

CompHysteresisHeating
7.00 K
ff0207 7000 3000 f000 10

CompHysteresisCooling
5.00 K
ff0207 5000 3000 f000 10

CompStartHeatingFrom
−60 °min
ff0225 c4ff 9cff e2ff 05

CompStartCoolingFrom
60 °min
ff0225 3c00 1e00 6400 05

MaxRemainingDeltaP
900
ff022d 8403 c800 8403 0a

MaximumS21BlockAllowed
5 h
ff020d 0500 0000 0900 01

NoiseReductionLevel
40 %
ff021b 2800 1e00 3c00 01

BlockTimeAfterRestart
0 min
ff0234 0000 0000 7800 01

BuildingCircuitPumpOutputHeating
auto
ff0240 3100 3100 6400 01

BuildingCircuitPumpOutputCooling
auto
ff0240 3100 3100 6400 01

BuildingCircuitPumpOutputHwc
100 %
ff0240 6400 3100 6400 01

HMUSystemStarts
23
ff0802 1700 0000 0000 00

RunStatsCoolHours = 35 h is a good independent check: this system only
started cooling at the end of August 2026 (the function was retrofitted),
so a small number is exactly right. Your 603 h and my 35 h are consistent
with two very different histories.

13 registers answer with a 2-byte refusal

All identical: NN=02, payload ff 01.

HeatCurve (352d), SummerSwitchOffTemp (352e), HcBivalencePoint (352f),
HwcBivalencePoint (3530), HcAlternativePoint (3531), MaxFlowTemp (3532),
MinFlowTemp (3533), HcEnabled (3534), HwcEnabled (3535),
EmergencyModeImmersionHeaterEnabled (3537), HwcChargeHysteresis (3539),
CoolTargetFlowTemp (352b), ImmersionHeaterEnabled (331e)

3108b51a0405ff352db6<0002ff0156>00

The b51a 05ff xxxx response layout

Comparing the two groups makes the format readable, which I think is worth
recording:

ff | <value LE 2B> | <min LE 2B> | <max LE 2B> | <step 1B>

   -

   flag = 02 → writable parameter, value block follows
   -

   flag = 08 → read-only value, min/max/step all zero (AirIntakeTemp,
   HMUSystemStarts)
   -

   flag = 01 → parameter not present on this device, reply ends after 2
   bytes

That explains the IGN:3 in the definitions, and it means the device hands
you min/max/step for free. Cross-check: CompHysteresisHeating reports min
3.00, max 15.00, step 1.00 — exactly the @minValue(3) / @maxValue(15) in
your definition.

I’d read the ff 01 group as “not implemented on this installation” rather
than a definition problem — so no change requested. My best guess is that
on a cascade the heating-curve and bivalence parameters live in the
controller rather than in the HMU, but that is a guess, not a measurement.

Where the device disagrees with the annotations

Register
Device reports
Definition

CompStartHeatingFrom
min −100, max −30
@minValue(-999) @maxValue(9)

CompStartCoolingFrom
min 30, max 100
@minValue(0) @maxValue(999)

NoiseReductionLevel
min 30, max 60
@minValue(40) @maxValue(60)

Only affects write validation. Interesting that NoiseReductionLevel differs
between our units: your @Ebus.example payload carries min 2800 = 40, mine
reports 1e00 = 30 — so that bound looks installation-dependent and may be
better left off.

One collision worth knowing about

ActiveCoolingEnabled = b51a 05ff331b is the same register I have been
carrying since 2026-08-29 under the name CoolingTechnology. I found it when
the VWZ (master 71) wrote b51a 06 331b 0100 to the HMU at the exact moment
the setting was changed at the indoor unit display. I read it as UIN with
0=none;1=active plus the min/max fields; you read the first byte as UCH
with 0=off;1=on. Same register, same meaning — just two names.

That only bit me because both lines were in one file during the test:

[main error] error reading scan config file vaillant/08.hmu.HW5103.csv for
ID “hmu00”,
SW0902, HW5103: ERR: duplicate entry, …:145: ERR: duplicate entry,
duplicate ID
[main error] scan config 08: ERR: duplicate entry

Not a problem for your PR as such, since CoolingTechnology is not upstream
— but if it ever goes upstream, one of the two names has to give way. Worth
noting that ebusd’s --checkconfig does not catch this; it counted all
messages happily and the collision only surfaced when the scan config was
loaded at runtime, aborting the rest of the file.

ClausLDK:
@buliwyf42  Correction to my earlier SW0905 Errorhistory test:

The empty responses I reported were caused by my micro-ebusd `hex` command including the telegram length byte (`NN=03`) as data.

Re-testing with the correct micro-ebusd syntax:

hex -s 31 08b503010100
hex -s 31 08b503010101
...
hex -s 31 08b503010109

all ten indices return a full 10-byte Errorhistory record on:

MF=Vaillant; ID=HMU00; SW=0905; HW=5103

Example:

3108b50303010102 / 0a010000000000ffffff00

All current slots are empty (`status=1`, `error=ffff`), so I cannot independently verify the BCD time/date encoding from a populated entry, but Errorhistory itself is definitely implemented on SW0905.

So my earlier conclusion about a SW0902/SW0905 difference was incorrect. Keeping the generic `Errors_inc` definition in #660 is the correct choice, and #662 remains the appropriate place for the BCD decoding fix.

buliwyf42:
Thanks for re-testing, and for posting the correction — that settles it from my side as well.

For the record on #662: the BCD evidence there rests on two independent sources — the populated slots on my HMU00/SW0902 (raw records earlier in this thread) and ctlv2/SW0514, which decodes the same way. Since all your slots are empty, an SW0905 cross-check of the encoding would need a populated entry, so there is nothing to chase unless one ever appears.

ClausLDK:
Thanks, agreed — #660 stays unchanged and #662 handles the BCD fix

Could you maybe approve my PRs (not sure if this will speed anything up @john30 ?), but it would be nice to have these done before autumn really kicks in!

buliwyf42 (approved):
Approving based on my hardware test on SW0902 earlier in this thread.

On your other open PRs (#655, #657, #659, #661): I haven't looked at those closely enough to approve them, and I'd rather not approve blind — but I'm happy to if I get the chance to test them. #670 is tested and approved.

